PR #2, opened 2016-10-11
use node cluster

rothfels:
Tested locally, opened 3 connections for 3 workspaces and saw them round-robin'ed across 3 separate node processes.

### DIFF
--- a/src/language-server.ts
+++ b/src/language-server.ts
@@ -1,19 +1,15 @@
 /// <reference path="../typings/node/node.d.ts"/>
 /// <reference path="../typings/vscode-extension-vscode/es6.d.ts"/>
 
-var net = require('net');
-var fs = require('fs');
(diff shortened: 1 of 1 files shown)


keegancsmith:
Just double checking, a workspace stays on one node process right due to the persistent connection between lsp-proxy and the LSP server?

rothfels:
Yep!
 





On Wed, Oct 12, 2016 3:01 AM, Keegan Carruthers-Smith notifications@github.com
wrote:
Just double checking, a workspace stays on one node process right due to the
persistent connection between lsp-proxy and the LSP server?

—
You are receiving this because you modified the open/close state.
Reply to this email directly, view it on GitHub, or mute the thread.



John Rothfels

The future sooner @ ✱Sourcegraph

(916) 990 4297

PR #8, opened 2016-10-19
WIP: faster workspace/symbol

beyang:


src/connection.ts
```diff
@@ -181,9 +181,15 @@ export default class Connection {
                         } else {
                             const navigateToItems = service.getWorkspaceSymbols(params.query, params.limit);
                             if (navigateToItems) {
+                                let sourceFiles = new Map<string, SourceFile>(); // cache SourceFiles
                                 result = navigateToItems.map(item => {
-                                    let start = ts.getLineAndCharacterOfPosition(service.services.getProgram().getSourceFile(item.fileName), item.textSpan.start);
-                                    let end = ts.getLineAndCharacterOfPosition(service.services.getProgram().getSourceFile(item.fileName), item.textSpan.start + item.textSpan.length);
+                                    let sourceFile = sourceFiles.get(item.fileName);
+                                    if (!sourceFile) {
+                                        sourceFile = service.services.getProgram().getSourceFile(item.fileName);
```
alexsaveliev: the bottleneck here is not `getSourceFile()` which is `Map::get()` but `getProgram()` that actualizes data. I think there is no need to back up source files with one more map, rather call `getProgram()` once

PR #15, opened 2016-10-25
Disable symbols request until we find a performant way to complete the request

nicot:


client/ts.go
```diff
@@ -157,6 +157,7 @@ func getReferences(client *jsonrpc2.Client) error {
 }
 
 func getWorkspaceSymbols(client *jsonrpc2.Client, query string, limit int) error {
+	return nil
```
beyang: Add a comment explaining why this line is here.

PR #17, opened 2016-10-26
Added tests

alexsaveliev:
- added unit tests 
- added Travis CI tests
- updated README.md


README.md
```diff
@@ -14,11 +14,38 @@ To try it in [Visual Studio Code](https://code.visualstudio.com), install the [v
 
 Run `node_modules/.bin/tsc --watch`.
 
+## Tests
+
+Run `node_modules/.bin/mocha build/test/*.js`
```
sqs: > Run `npm test` 

is more standard/simpler

sqs: or `yarn test` :)

src/test/test-utils.ts
```diff
@@ -0,0 +1,214 @@
...
+            channel.serverConnection.start();
+            channel.clientConnection.start();
+
+            const params: vscode.InitializeParams = {
+                processId: null,
+                rootPath: 'file:///',
+                capabilities: null
+            }
+
+            channel.clientConnection.sendRequest(rt.InitializeRequest.type, params).then(function () {
```
sqs: This applies to the whole codebase, but why not use ES6 `() => ...` syntax? `function() { ... }` is much more verbose and has worse handling for `this` in most cases.

tsd.json
```diff
@@ -7,6 +7,15 @@
   "installed": {
```
sqs: why this instead of `npm install --save-dev @types/whatever`?

src/test/test-utils.ts
alexsaveliev: bad habit, sorry, will take into account in a separate PR :)

PR #20, opened 2016-10-27
D.ts support

antonina-cherednichenko:


src/project-manager.ts
```diff
@@ -483,6 +536,35 @@ export class ProjectConfiguration {
...
+                            let res = self.findDTSFile(dep);
+                            if (res && res.length > 0) {
+                                let depFile = res[0];
+                                self.fs.addFile(`node_modules/@types/${dep}/index.d.ts`, fs.readFileSync(res[0], 'utf8'));
+                                self.dtsNames[dep] = path_.basename(depFile);
+                            }
+                        }
+                    }
+
+                    if (devDeps) {
```
alexsaveliev: Please extract method and use it to process dependencies and devDependencies

src/project-manager.ts
```diff
@@ -483,6 +536,35 @@ export class ProjectConfiguration {
...
+                        for (let devDep in devDeps) {
+                            let res = self.findDTSFile(devDep);
+                            if (res && res.length > 0) {
+                                let depFile = res[0];
+                                self.fs.addFile(`node_modules/@types/${devDep}/index.d.ts`, fs.readFileSync(depFile, 'utf8'));
```
alexsaveliev: Shouldn't we rater wait until TypeScript will ask our `LanguageServiceHost` for a file from `.../node_modules/@types/foo/name.d.ts` and only then we'll try to read local file and add it to in-memory FS under the name of `.../node_modules/@types/foo/name.d.ts`? This will solve the following:
- what if TypeScript will ask for `foo/bar.d.ts` instead of `foo/index.d.ts`?
- what if package.json is not located in the workspace root? (foo/bar/package.json)

src/typescript-service.ts
```diff
@@ -32,11 +34,24 @@ export default class TypeScriptService {
...
         this.projectManager = new pm.ProjectManager(root, strict, connection);
 
+        this.fetchDefinetelyTypedRepo();
+
         //initialize providers 
         this.exportedSymbolProvider = new ExportedSymbolsProvider(this);
         this.externalRefsProvider = new ExternalRefsProvider(this);
     }
 
+    fetchDefinetelyTypedRepo() {
```
alexsaveliev: typo

src/typescript-service.ts
```diff
@@ -104,12 +119,26 @@ export default class TypeScriptService {
...
-                                    end: end
-                                }));
+
+                                let pathParts = def.fileName.split(path.sep);
+                                let index1 = pathParts.indexOf("node_modules")
+                                let index2 = pathParts.indexOf("@types");
+                                if (index2 - index1 == 1) {
+                                    let dtsFolder = pathParts[index2 + 1];
+                                    let dtsFileName = configuration.dtsNames[dtsFolder];
+                                    ret.push(Location.create(`https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/${dtsFolder}/${dtsFileName}`, {
```
alexsaveliev: Are you in sync with @sqs regarding URL format? (Just checking..)

src/project-manager.ts
antonina-cherednichenko: done

src/typescript-service.ts
antonina-cherednichenko: done

src/typescript-service.ts
```diff
@@ -32,11 +34,24 @@ export default class TypeScriptService {
...
+        this.fetchDefinetelyTypedRepo();
+
         //initialize providers 
         this.exportedSymbolProvider = new ExportedSymbolsProvider(this);
         this.externalRefsProvider = new ExternalRefsProvider(this);
     }
 
+    fetchDefinetelyTypedRepo() {
+        //TODO add checking that repo is already fetched
+        github('DefinitelyTyped/DefinitelyTyped', '/tmp/DefinitelyTyped', function (err) {
```
antonina-cherednichenko: command works asynchronously, we won't wait for end of downloading

src/typescript-service.ts
```diff
@@ -104,12 +119,26 @@ export default class TypeScriptService {
                                 const sourceFile = configuration.program.getSourceFile(def.fileName);
                                 const start = ts.getLineAndCharacterOfPosition(sourceFile, def.textSpan.start);
                                 const end = ts.getLineAndCharacterOfPosition(sourceFile, def.textSpan.start + def.textSpan.length);
-                                ret.push(Location.create(util.path2uri(self.root, def.fileName), {
-                                    start: start,
-                                    end: end
-                                }));
+
+                                let pathParts = def.fileName.split(path.sep);
```
antonina-cherednichenko: https://nodejs.org/api/path.html#path_path_sep, should be platform independent

src/typescript-service.ts
antonina-cherednichenko: Will discuss today

src/project-manager.ts
```diff
@@ -446,16 +476,39 @@ export class ProjectConfiguration {
...
+
+    public setPackageJsonFile(packageJsonFileName: string) {
+        this.packageJsonFileName = packageJsonFileName;
+    }
+
+    public findDTSFile(dep): string[] {
+        let jsFiles = findfiles(`/tmp/DefinitelyTyped/${dep}`, {
+            exclude: [],
+            matcher: function (directory, name) {
+                return (/d\.ts$/i).test(name);
```
antonina-cherednichenko: Done

src/project-manager.ts
```diff
@@ -446,16 +476,39 @@ export class ProjectConfiguration {
...
         this.configContent = configContent;
+        this.packageJsonFileName = packageJsonFileName;
+        this.dtsNames = {};
+    }
+
+    public setPackageJsonFile(packageJsonFileName: string) {
+        this.packageJsonFileName = packageJsonFileName;
+    }
+
+    public findDTSFile(dep): string[] {
```
antonina-cherednichenko: Done

src/typescript-service.ts
```diff
@@ -32,11 +34,24 @@ export default class TypeScriptService {
...
 
+        this.fetchDefinetelyTypedRepo();
+
         //initialize providers 
         this.exportedSymbolProvider = new ExportedSymbolsProvider(this);
         this.externalRefsProvider = new ExternalRefsProvider(this);
     }
 
+    fetchDefinetelyTypedRepo() {
+        //TODO add checking that repo is already fetched
```
antonina-cherednichenko: Done

src/typescript-service.ts
```diff
@@ -32,11 +34,30 @@ export default class TypeScriptService {
...
         //initialize providers 
         this.exportedSymbolProvider = new ExportedSymbolsProvider(this);
         this.externalRefsProvider = new ExternalRefsProvider(this);
     }
 
+    /**
+     * Fetch repository from https://github.com/DefinitelyTyped/DefinitelyTyped into /tmp/DefinitelyTyped/
+     */
+    fetchDefinitelyTypedRepo() {
+        fs.exists('/tmp/DefinitelyTyped', (exists) => {
```
sqs: If the clone fails, then will subsequent operations attempt to reclone? Cloning from GitHub probably fails 2% of the time, and it would be bad if the clone failed when this server first started up and we have no way of knowing (other than checking elevated error rates) or recovering.

Will an error be returned that can be associated with the particular session that it occurred in, or will we have to look at the server logs? (Seems like the latter.)

src/typescript-service.ts
```diff
@@ -104,12 +125,26 @@ export default class TypeScriptService {
...
-                                    end: end
-                                }));
+
+                                let pathParts = def.fileName.split(path.sep);
+                                let index1 = pathParts.indexOf("node_modules")
+                                let index2 = pathParts.indexOf("@types");
+                                if (index2 - index1 == 1) {
+                                    let dtsFolder = pathParts[index2 + 1];
+                                    let dtsFileName = configuration.dtsNames[dtsFolder];
+                                    ret.push(Location.create(`git://github.com/DefinitelyTyped/DefinitelyTyped?master#${dtsFolder}/${dtsFileName}`, {
```
sqs: Hardcode the commit ID for determinism.

src/project-manager.ts
```diff
@@ -446,16 +475,57 @@ export class ProjectConfiguration {
...
+    }
+
+    /**
+     * Finds d.ts files located in /tmp/DefinitelyTyped folder which are related to dependencies from package.json
+     * @param deps list of dependecies from package.json file
+     */
+    public processDeps(deps) {
+        if (deps) {
+            for (let dep in deps) {
+                let findRes = findfiles(`/tmp/DefinitelyTyped/${dep}`, {
```
sqs: What type is `deps`? Is it an array or object? Can you add a TypeScript type? Also, `for (let x in y) { ... }` does not have consistent ordering for arrays, so it is not recommended to use it (see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...in).

sqs: If `dep` is `../../`, will this take a really long time and block while the whole file system tree is traversed?

E.g., if someone has code like `require('../../')`

src/project-manager.ts
```diff
@@ -483,6 +553,20 @@ export class ProjectConfiguration {
                 if (/(^|\/)jsconfig\.json$/.test(self.configFileName)) {
                     options.allowJs = true;
                 }
+
+                if (self.packageJsonFileName) {
+                    try {
+                        let parsedResult = JSON.parse(self.fs.readFile(self.packageJsonFileName));
```
sqs: Do we santiize `self.packageJsonFileName` so it can't be `/etc/passwd` for example?

src/project-manager.ts
antonina-cherednichenko: Traversing properties of object, order is not important here
These are deps from package.json dependencies section, second part of question is not clear to me

src/typescript-service.ts
antonina-cherednichenko: If code will be moved from js/ts server itself, to its installation script, re-cloning can be scheduled

sqs:
The reason why I feel it belongs at this level is that when we support actually fetching NPM modules, we obviously can't prefetch them all. So if something depends on minimist@1.2.3 then we will have to fetch it here. We need to support efficiently fetching, retrying, reusing anyway for that. 

Also, is it possible that instead of fetching the whole DT git repo, we could just fetch the (much smaller) NPM .tar.gz for the @types/whatever module that is needed? That would make this problem more like the problem I describe above (which we need to solve anyway).

src/typescript-service.ts
antonina-cherednichenko: We don't need to fetch node_modules anymore and we actually are not fetching them. All we need is to download DefinitelyTyped repo somewhere, this repo will be shared for all processes (claster mode), all users etc. And update it periodically since people updating d.ts files (once a day is enough) or while restarting of sourcegraph. Size of repo is not so big 128 Mb. 
Maybe it would be easier if I explain it this way: imagine that these d.ts files are source code files of langserver project located in third part repo - DefinitelyTyped/DefinitelyTyped repo.

sqs:
Can you add some tests for the typings resolution - e.g., that if you have a workspace that imports something from DT that the typings get fetched properly and hover shows the right info?

PR #22, opened 2016-10-28
Use a lambda function for maybeEnough

keegancsmith:
The current version fails quite often due to `this` changing. Using a lambda
function gives us lexical scoping (I think).


keegancsmith:
Did a bunch of testing, this def fixes a bug. Will merge in now, but please revert if it looks wrong.

PR #24, opened 2016-11-03
Adding default library file content

alexsaveliev:
- adding default library's (lib.d.ts/lib.es6.d.ts) content read from the local file system. fixes sourcegraph/sourcegraph#2052
- ignoring `d.ts` files when searching for workspace symbols with not-empty query; excluding default library when searching for workspace symbols with an empty query


sqs:
@rothfels or @keegancsmith can one of you please review instead?

PR #26, opened 2016-11-08
lower initialization cost & memory footprint for definition, hover, workspace/symbol, references

beyang:
This decreases the initialization time for the language server and lowers the memory footprint for most LSP operations (with the exception of `references` on tokens in dependency files; the memory footprint here stays the same).


alexsaveliev:
Could you please reformat the code to use 4 spaces instead of tabs?

src/project-manager.ts
```diff
@@ -134,55 +109,114 @@ export class ProjectManager {
...
+    // project exists in localFs. TypeScript/JavaScript module
+    // structure is determined by [jt]sconfig.json, filesystem layout,
+    // global*.d.ts files. For performance reasons, we only read in
+    // the contents of some files and store "var placeholder;" as the
+    // contents of all other files.
+    ensureModuleStructure(): Promise<void> {
+        if (!this.ensuredModuleStructure) {
+            this.ensuredModuleStructure = this.ensureModuleStructure_();
+            this.ensuredModuleStructure.catch((err) => {
+                this.ensuredModuleStructure = null;
```
alexsaveliev: Shouldn't we display an error?

src/project-manager.ts
```diff
@@ -134,55 +109,114 @@ export class ProjectManager {
...
-        let counter: number = 0;
-
-        let cb = (err: Error, result?: FileSystem.FileInfo[]) => {
+        const self = this;
+        const filesToFetch = [];
+        return this.walkRemote(this.root, function(path: string, info: FileSystem.FileInfo, err?: Error): (Error | null) {
             if (err) {
-                console.error('got error while reading dir', err);
-                return callback(err)
+                return err ? err : null;
```
alexsaveliev: Useless ternary operator

src/typescript-service.ts
```diff
@@ -83,13 +84,161 @@ export default class TypeScriptService {
...
+            console.error("Failed to fetch files for workspace/symbol:", err);
+            this.ensuredFilesForWorkspaceSymbol = null;
+        });
+
+        return promise;
+    }
+
+    private ensuredAllFiles: Promise<void> = null;
+
+    private ensureFilesForReferences(uri: string): Promise<void> {
```
alexsaveliev: What's the difference between `...ForWorkspaceSymbols` and `...ForReferences`? Shouldn't we use the same logic for both?

beyang: There is one difference. `ensureFilesForReferences` fetches all files, including those in `node_modules` if the referenced token is in a `node_modules` file. It does so, because references to those tokens could be defined in any file in the repository. It avoids doing this when it doesn't need to, because `node_modules` is often huge.

alexsaveliev: Okay

src/typescript-service.ts
```diff
@@ -83,13 +84,161 @@ export default class TypeScriptService {
...
+        }).then(() => {
+            return this.projectManager.refreshConfigurations();
+        }).then(() => {
+            return this.projectManager.getConfiguration(fileName).get().then((config) => {
+                const contents = config.moduleResolutionHost().readFile(fileName);
+                const info = ts.preProcessFile(contents, true, true);
+                const compilerOpt = config.host.getCompilationSettings();
+
+                const importDirs = new Set<string>();
+                for (const imp of info.importedFiles) {
```
alexsaveliev: We should test the following case:
A imports B, B imports C. A.a is an instance of B.foo which is {bar: string; baz: C.qux}
What we'll see as A.a's hover if C is not in the `info.importedFiles`?

src/project-manager.ts
```diff
@@ -134,55 +109,115 @@ export class ProjectManager {
...
-            let tasks = [];
-            result.forEach((fi) => {
-                if (fi.name.indexOf('/.') >= 0) {
-                    return
+            const rel = path_.posix.relative(self.root, path);
+            if (util.isGlobalTSFile(rel) || util.isConfigFile(rel) || util.isPackageJsonFile(rel)) {
+                filesToFetch.push(path)
+            } else {
+                if (!self.localFs.fileExists(rel)) {
+                    self.localFs.addFile(rel, "var dummy_0ff1bd;");
```
beyang: this is never the case, due to the outer if conditional

beyang:
(Merged)

PR #29, opened 2016-11-15
Performance optimization and bug fixes

alexsaveliev:
- there is no need to synchronize configuration when looking for hovers and definitions. Files required to produce results should be already fetched and parsed while processing main file and its refs and imports. `ProjectManager::syncConfiguration` leads to parsing the whole workpace and required for references and workspace symbols only
- processing triple slash refs when collecting dependents, partially addresses sourcegraph/sourcegraph#2201
- do not throw an error when we were unable to fetch file from VFS
- added debug facilities
- fetching all TypeScript library files from local file system. Addresses sourcegraph/sourcegraph#2127 and sourcegraph/sourcegraph#2249

@beyang It's unclear why do we limit depth to 3 while fetching imports and dependencies? This particular issue provides a case when hover/definition information is not available because we haven't fetched all the required dependencies..

### DIFF
--- a/src/project-manager.ts
+++ b/src/project-manager.ts
@@ -388,9 +388,10 @@ class InMemoryLanguageServiceHost implements ts.LanguageServiceHost {
         this.versions = versions;
         this.projectVersion = 1;
         this.files = [];
-        // adding content of default library file read from the local file system 
-        const defaultLib = util.normalizePath(this.getDefaultLibFileName(options));
-        this.fs.entries[defaultLib] = fs_.readFileSync(defaultLib).toString();
+        // adding library files from the local file system
+        readTsLibraries().forEach((content, name) => {
+            this.fs.entries[name] = content;
+        });
     }
 
     /**
@@ -477,6 +478,7 @@ export class InMemoryFileSystem implements ts.ParseConfigHost, ts.ModuleResoluti
     entries: any;
     overlay: any;
 
+
     useCaseSensitiveFileNames: boolean;
 
     path: string;
@@ -505,18 +507,7 @@ export class InMemoryFileSystem implements ts.ParseConfigHost, ts.ModuleResoluti
     }
 
     fileExists(path: string): boolean {
-        if (!!this.overlay[path]) {
-            return true;
-        }
-        const rel = path_.posix.relative('/', path);
-        if (!!this.overlay[rel]) {
-            return true;
-        }
-
-        if (!!this.entries[path]) {
-            return !!this.entries[path];
-        }
-        return !!this.entries[rel];
+        return this.readFile(path) !== undefined;
     }
 
     readFile(path: string): string {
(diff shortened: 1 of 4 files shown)


src/project-manager.ts
```diff
@@ -382,9 +388,10 @@ class InMemoryLanguageServiceHost implements ts.LanguageServiceHost {
         this.versions = versions;
         this.projectVersion = 1;
         this.files = [];
-        // adding content of default library file read from the local file system 
-        const defaultLib = util.normalizePath(this.getDefaultLibFileName(options));
-        this.fs.entries[defaultLib] = fs_.readFileSync(defaultLib).toString();
+        // adding library files from the local file system
+        readTsLibraries().forEach((content, name) => {
```
beyang: Actually, it's quite strange to me that this works. The TS library paths are all absolute on my local machine, so this ends up adding a bunch of files to the in-memory FS that begin with `/home/beyang/...`. Somehow, the TypeScript compiler fetches these file paths from the local FS, even though it should be agnostic about the local FS when talking to the in-memory language host.

src/test/language-server-test.ts
```diff
@@ -346,4 +346,176 @@ describe('LSP', function () {
...
+                            character: 4
+                        },
+                        end: {
+                            line: 3,
+                            character: 5
+                        }
+                    }
+                }, done);
+        });
+        it('TODO deep definition', function (done: (err?: Error) => void) {
```
beyang: remove "TODO"?

src/test/language-server-test.ts
```diff
@@ -346,4 +346,176 @@ describe('LSP', function () {
...
+                    utils.definition({
+                        textDocument: {
+                            uri: 'file:///deeprefs/a.ts'
+                        },
+                        position: {
+                            line: 2,
+                            character: 0
+                        }
+                    }, null /*{
+                            uri: 'file:///deeprefs/d.ts',
```
beyang: Why the commented-out section?

src/test/language-server-test.ts
```diff
@@ -346,4 +346,176 @@ describe('LSP', function () {
...
+            utils.setUp({
+                'tsconfig.json': JSON.stringify({
+                    compilerOptions: {
+                        lib: ['es2016', 'dom']
+                    }
+                }),
+                'a.ts': "function foo(n: Node): {console.log(n.parentNode})}"
+            }, done);
+        });
+        it('thou shall load local library file', function (done: (err?: Error) => void) {
```
beyang: `s/thou shall load local library file/loads local library file/`

src/typescript-service.ts
```diff
@@ -270,8 +284,6 @@ export default class TypeScriptService {
                 const configuration = this.projectManager.getConfiguration(fileName);
                 configuration.prepare(this.connection).then(() => {
                     try {
-                        this.projectManager.syncConfiguration(configuration, this.connection); // resync after getting new config
```
beyang: Isn't this necessary, because some files might have previously held the placeholder content?

src/test/language-server-test.ts
alexsaveliev: This test passes but should fail because we do not support definition resolution for depths > 3 (A refers to B, B refers to C, C refers to D, D defines something). There are samples in microsoft/typescript code. I'd like to remove TODO when we'll deal what to do with depth limitation

src/project-manager.ts
alexsaveliev: We are fetching content of libraries TS compiler carries with itself. See "lib" here: http://json.schemastore.org/tsconfig. We do not read content of local file system outside our `node_modules/typescript/*.d.ts`

src/typescript-service.ts
alexsaveliev: we won't touch placeholders because before parsing files we'll ensure they are fetched from VFS and will operate with the real data

PR #32, opened 2016-11-16
Definitions from TypeScript libraries

alexsaveliev:
- resolving TS library definitions to github.com/Microsoft/TypeScript#lib/..., addresses sourcegraph/sourcegraph#2248
- excluding symbols defined in TS libraries from workspace symbols
- fixed and enhanced tests

Tests are failing because they depend on #31

### DIFF
--- a/src/project-manager.ts
+++ b/src/project-manager.ts
@@ -389,7 +389,7 @@ class InMemoryLanguageServiceHost implements ts.LanguageServiceHost {
         this.projectVersion = 1;
         this.files = [];
         // adding library files from the local file system
-        readTsLibraries().forEach((content, name) => {
+        getTypeScriptLibraries().forEach((content, name) => {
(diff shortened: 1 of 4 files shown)


beyang:
@alexsaveliev I merged this, because it looks good, but can you please respond to comments?

alexsaveliev:
> Now that this change is, should the previous change of serving non-existent files as files with empty contents be reverted? Or is that still necessary?

Still necessary, for example when there is a missing triple slash reference. Missing file should not prevent parsing the rest of files

> Why get rid of the old test cases and replace them with new ones instead of just adding new test cases?

Well, it turned out that there was a missing error handler so test silently failed. I have added error handler to report failures and then changed test case to be valid TS code (you can use un-exported definitions from references or exported ones from imports in the same file but not both).

PR #34, opened 2016-11-24
switching to yarn

alexsaveliev:
- using `yarn` to build and test project
- removed unused code and dependencies


.travis.yml
```diff
@@ -3,9 +3,12 @@ language: node_js
 node_js:
   - "6"
 
+before_install:
+  - npm install yarn -g
```
keegancsmith: This is currently the suggested way to get yarn on travis https://github.com/travis-ci/travis-ci/issues/6720#issuecomment-254988120

That mixed with some official support for yarn caching should make our CI super fast https://blog.travis-ci.com/2016-11-21-travis-ci-now-supports-yarn

.travis.yml
```diff
@@ -3,9 +3,12 @@ language: node_js
...
+
 install:
-  - npm install
+  - yarn install
```
keegancsmith: FYI you can just do `yarn` without the install args

.travis.yml
alexsaveliev: it seems that yarn is already available without the need to install anything. `.travis.yml` script works without the need to specicy `before_install` and `install` steps. I have added `cache` as suggested:

```yml
cache:
  yarn: true
  directories:
    - node_modules
```

alexsaveliev:
>  But why all the deleted code? 

There was a dependency to mine GitHub repository and code that used it wasn't used/optimized

> Are you not deleting some of the perf/customizations we have done?

Nope, I have deleted magic workspace/symbol methods - "externals" and "exported"

keegancsmith:
> Nope, I have deleted magic workspace/symbol methods - "externals" and "exported"

We may actually want those. cc @slimsag @beyang. Not a big deal at the moment since:
* I don't think this code path is currently used in production
* The spec we have for that behaviour may have changed anyways
* We can easily recover the code.

PR #35, opened 2016-11-29
using TS version constant in TS library links

alexsaveliev:
- when forming links to TS library, using `ts.version` constant to keep links up to date with the TS version being used. Followup for sourcegraph/sourcegraph#2372 and @d8cf3b279b3826c88df1b2fe67131128197fc6aa


beyang:
LGTM, thanks!

PR #36, opened 2016-12-01
workspace/reference and textDocument/documentSymbol

beyang:
Adds `workspace/reference` and `textDocument/documentSymbol` endpoints.

### DIFF
--- a/src/connection.ts
+++ b/src/connection.ts
@@ -118,14 +118,13 @@ export default class Connection {
 		this.connection.onRequest(rt.WorkspaceSymbolsRequest.type, (params: rt.WorkspaceSymbolParamsWithLimit): Promise<SymbolInformation[]> => {
 			const enter = new Date().getTime();
 			return new Promise<SymbolInformation[]>((resolve, reject) => {
-				let result = [];
 				const init = new Date().getTime();
 				try {
 					return service.getWorkspaceSymbols(params.query, params.limit).then((result) => {
 						result = result ? result : [];
 						const exit = new Date().getTime();
 						console.error('workspace/symbol', params.query, 'total', (exit - enter) / 1000.0, 'busy', (exit - init) / 1000.0, 'wait', (init - enter) / 1000.0);
-						return resolve(result);
+						return resolve(result || []);
 					});
 				} catch (e) {
 					console.error(params, e);
@@ -137,14 +136,13 @@ export default class Connection {
 		this.connection.onRequest(rt.DocumentSymbolRequest.type, (params: DocumentSymbolParams): Promise<SymbolInformation[]> => {
 			const enter = new Date().getTime();
 			return new Promise<SymbolInformation[]>((resolve, reject) => {
-				let result = [];
 				const init = new Date().getTime();
 				try {
 					return service.getDocumentSymbol(params).then((result) => {
 						result = result ? result : [];
 						const exit = new Date().getTime();
 						console.error('textDocument/documentSymbol', "", 'total', (exit - enter) / 1000.0, 'busy', (exit - init) / 1000.0, 'wait', (init - enter) / 1000.0);
-						return resolve(result);
+						return resolve(result || []);
 					});
 				} catch (e) {
 					console.error(params, e);
@@ -153,9 +151,9 @@ export default class Connection {
 			});
 		});
 
-		this.connection.onRequest(rt.WorkspaceReferenceRequest.type, (params: util.WorkspaceReferenceParams): Promise<util.ReferenceInformation[]> => {
+		this.connection.onRequest(rt.WorkspaceReferenceRequest.type, (params: rt.WorkspaceReferenceParams): Promise<rt.ReferenceInformation[]> => {
 			const enter = new Date().getTime();
-			return new Promise<util.ReferenceInformation[]>((resolve, reject) => {
+			return new Promise<rt.ReferenceInformation[]>((resolve, reject) => {
 				let result = [];
 				const init = new Date().getTime();
 				try {
--- a/src/connection.ts
+++ b/src/connection.ts
@@ -12,6 +12,7 @@ import {
 	Location,
 	Hover,
 	WorkspaceSymbolParams,
+	DocumentSymbolParams,
 	SymbolInformation,
 	RequestType,
 	Range,
@@ -123,7 +124,26 @@ export default class Connection {
 					return service.getWorkspaceSymbols(params.query, params.limit).then((result) => {
 						result = result ? result : [];
 						const exit = new Date().getTime();
-						console.error('symbol', params.query, 'total', (exit - enter) / 1000.0, 'busy', (exit - init) / 1000.0, 'wait', (init - enter) / 1000.0);
+						console.error('workspace/symbol', params.query, 'total', (exit - enter) / 1000.0, 'busy', (exit - init) / 1000.0, 'wait', (init - enter) / 1000.0);
+						return resolve(result);
+					});
+				} catch (e) {
+					console.error(params, e);
+					return resolve([]);
+				}
+			});
+		});
+
+		this.connection.onRequest(rt.DocumentSymbolRequest.type, (params: DocumentSymbolParams): Promise<SymbolInformation[]> => {
+			const enter = new Date().getTime();
+			return new Promise<SymbolInformation[]>((resolve, reject) => {
+				let result = [];
+				const init = new Date().getTime();
+				try {
+					return service.getDocumentSymbol(params).then((result) => {
+						result = result ? result : [];
+						const exit = new Date().getTime();
+						console.error('textDocument/documentSymbol', "", 'total', (exit - enter) / 1000.0, 'busy', (exit - init) / 1000.0, 'wait', (init - enter) / 1000.0);
 						return resolve(result);
 					});
 				} catch (e) {
--- a/src/connection.ts
+++ b/src/connection.ts
@@ -133,6 +133,26 @@ export default class Connection {
 			});
 		});
 
+		this.connection.onRequest(rt.WorkspaceReferenceRequest.type, (params: util.WorkspaceReferenceParams): Promise<util.ReferenceInformation[]> => {
+			const enter = new Date().getTime();
+			return new Promise<util.ReferenceInformation[]>((resolve, reject) => {
+				let result = [];
+				const init = new Date().getTime();
+				try {
+					return service.getWorkspaceReference(params).then((result) => {
+						result = result ? result : [];
+						const exit = new Date().getTime();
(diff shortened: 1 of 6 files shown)


src/connection.ts
```diff
@@ -133,6 +134,45 @@ export default class Connection {
 			});
 		});
 
+		this.connection.onRequest(rt.DocumentSymbolRequest.type, (params: DocumentSymbolParams): Promise<SymbolInformation[]> => {
+			const enter = new Date().getTime();
+			return new Promise<SymbolInformation[]>((resolve, reject) => {
+				let result = [];
```
alexsaveliev: redundant

src/connection.ts
```diff
@@ -133,6 +134,45 @@ export default class Connection {
...
+				const init = new Date().getTime();
+				try {
+					return service.getDocumentSymbol(params).then((result) => {
+						result = result ? result : [];
+						const exit = new Date().getTime();
+						console.error('textDocument/documentSymbol', "", 'total', (exit - enter) / 1000.0, 'busy', (exit - init) / 1000.0, 'wait', (init - enter) / 1000.0);
+						return resolve(result);
```
alexsaveliev: I'd do: `resolve(result || [])`

src/connection.ts
```diff
@@ -133,6 +134,45 @@ export default class Connection {
...
+					console.error(params, e);
+					return resolve([]);
+				}
+			});
+		});
+
+		this.connection.onRequest(rt.WorkspaceReferenceRequest.type, (params: util.WorkspaceReferenceParams): Promise<util.ReferenceInformation[]> => {
+			const enter = new Date().getTime();
+			return new Promise<util.ReferenceInformation[]>((resolve, reject) => {
+				let result = [];
```
alexsaveliev: redundant

src/connection.ts
```diff
@@ -133,6 +134,45 @@ export default class Connection {
...
+				const init = new Date().getTime();
+				try {
+					return service.getWorkspaceReference(params).then((result) => {
+						result = result ? result : [];
+						const exit = new Date().getTime();
+						console.error('workspace/reference', 'total', (exit - enter) / 1000.0, 'busy', (exit - init) / 1000.0, 'wait', (init - enter) / 1000.0);
+						return resolve(result);
```
alexsaveliev: I'd do: `resolve(result || [])`

src/test/test-utils.ts
```diff
@@ -205,6 +216,16 @@ export function symbols(params: rt.WorkspaceSymbolParamsWithLimit, expected: vsc
 	})
 }
 
+export function documentSymbols(params: vscode.DocumentSymbolParams, expected: vscode.SymbolInformation[], done: (err?: Error) => void) {
+	channel.clientConnection.sendRequest(rt.DocumentSymbolRequest.type, params).then((result: vscode.SymbolInformation[]) => {
+		check(done, () => {
+			chai.expect(result).to.deep.equal(expected);
+		});
+	}, (err?: Error) => {
+		return done(err || new Error('workspace symbols request failed'))
```
alexsaveliev: s/workspace/document

src/util.ts
```diff
@@ -225,3 +225,22 @@ let globalTSPattern = /(^|\/)globals?\.d\.ts$/;
 export function isGlobalTSFile(filename: string): boolean {
 	return globalTSPattern.test(filename);
 }
+
+
+/*
+ * WorkspaceReferenceParams holds parameters for the
```
alexsaveliev: I'd move them to `request-type.ts`, where similar `WorkspaceSymbolParamsWithLimit` is

src/typescript-service.ts
```diff
@@ -338,6 +338,847 @@ export default class TypeScriptService {
 		});
 	}
 
+	getDocumentSymbol(params: DocumentSymbolParams): Promise<SymbolInformation[]> {
+		const uri = params.textDocument.uri;
+		return this.ensureFilesForHoverAndDefinition(uri).then(() => {
+			const fileName = util.uri2path(uri);
+			const config = this.projectManager.getConfiguration(uri);
+			return config.prepare(this.connection).then(() => {
+				this.projectManager.syncConfiguration(config, this.connection);
```
alexsaveliev: I think that `ensureFilesForHoverAndDefinition` and `this.projectManager.syncConfiguration` do not match. You should use either first or second but not both. I's say that `syncConfiguration` is unnecessary here

src/typescript-service.ts
```diff
@@ -338,6 +338,847 @@ export default class TypeScriptService {
...
+				const sourceFile = config.program.getSourceFile(fileName);
+				const tree = config.service.getNavigationTree(fileName);
+				const limit = 100000;
```
alexsaveliev: I'd move this constant to the file's top

src/typescript-service.ts
```diff
@@ -338,6 +338,847 @@ export default class TypeScriptService {
...
+	}
+
+	getWorkspaceReference(params: util.WorkspaceReferenceParams): Promise<util.ReferenceInformation[]> {
+		const refInfo: util.ReferenceInformation[] = [];
+		return this.ensureFilesForWorkspaceSymbol().then(() => {
+			return Promise.all(this.projectManager.getConfigurations().map((config) => {
+				return config.prepare(this.connection).then((config) => {
+					this.projectManager.syncConfiguration(config, this.connection);
+					for (let source of config.service.getProgram().getSourceFiles()) {
+						if (source.fileName.indexOf("/node_modules/") !== -1) {
```
alexsaveliev: For a safety I'd do:
`if (util.normalizePath(source.fileName).indexOf(`${path_.posix.sep}node_modules${path_.posix.sep}`) !== -1) `

src/typescript-service.ts
```diff
@@ -338,6 +338,847 @@ export default class TypeScriptService {
...
+							continue;
+						}
+						this.walkMostAST(source, (node) => {
+							switch (node.kind) {
+								case ts.SyntaxKind.Identifier: {
+									const fileName = source.fileName.startsWith("/") ? source.fileName.substr(1) : source.fileName;
```
alexsaveliev: Will it work in non-strict mode on Windows?

src/typescript-service.ts
```diff
@@ -338,6 +338,847 @@ export default class TypeScriptService {
...
+			}));
+		}).then(() => {
+			return refInfo;
+		});
+	}
+
+	/*
+	 * walkMostAST walks most of the AST (the part that matters for gathering all references)
+	 */
+	walkMostAST(node: ts.Node, visit: (node: ts.Node) => void) {
```
alexsaveliev: make `private`?

src/typescript-service.ts
```diff
@@ -338,6 +338,847 @@ export default class TypeScriptService {
...
+		visit(node);
+		const children: ts.Node[] = [];
```
alexsaveliev: Do we need `children` array? Can't we visit all the children in place?

src/typescript-service.ts
```diff
@@ -338,6 +338,847 @@ export default class TypeScriptService {
...
+									const id = node as ts.Identifier;
+									const defs = config.service.getDefinitionAtPosition(fileName, node.pos + 1);
+
+									if (defs && defs.length > 0) {
+										const def = defs[0];
+										const start = ts.getLineAndCharacterOfPosition(source, node.pos);
+										const end = ts.getLineAndCharacterOfPosition(source, node.end);
+										const ref = {
+											location: {
+												uri: util.path2uri('', source.fileName),
```
alexsaveliev: use `this.defUri` instead to deal with TS-originated defs

src/typescript-service.ts
```diff
@@ -338,6 +338,847 @@ export default class TypeScriptService {
...
+												range: {
+													start: start,
+													end: end,
+												},
+											},
+											name: def.name,
+											containerName: def.containerName,
+											uri: util.path2uri('', def.fileName),
```
alexsaveliev: use `this.defUri` instead to deal with TS-originated defs. Also, please find an extract from results on Windows in non-strict mode (from vscode):
```
  { location:
     { uri: 'file:///c:/projects/langserver-typescript/src/util.ts',
       range: [Object] },
    name: 'ts',
    containerName: '',
    uri: 'file:///c:/projects/langserver-typescript/src/util.ts' },
```
Note uris here

src/typescript-service.ts
```diff
@@ -338,6 +338,847 @@ export default class TypeScriptService {
...
+		return this.ensureFilesForHoverAndDefinition(uri).then(() => {
+			const fileName = util.uri2path(uri);
+			const config = this.projectManager.getConfiguration(uri);
+			return config.prepare(this.connection).then(() => {
+				this.projectManager.syncConfiguration(config, this.connection);
+				const sourceFile = config.program.getSourceFile(fileName);
+				const tree = config.service.getNavigationTree(fileName);
+				const limit = 100000;
+				const result: SymbolInformation[] = [];
+				this.flattenNavigationTreeItem(tree, null, sourceFile, result, limit);
```
alexsaveliev: When tried to run in on Windows in non-strict mode (from vscode) I got the following items (extract):
```
  { name: 'TypeScriptService',
    kind: 15,
    location:
     { uri: 'file:///c:/projects/langserver-typescript/src/connection.ts',
       range: [Object] } },
  { name: 'util',
    kind: 15,
    location:
     { uri: 'file:///c:/projects/langserver-typescript/src/connection.ts',
       range: [Object] } },
```
Note absolute uris, we should make sure we are using relative ones

src/typescript-service.ts
beyang: We could do it in place, but it makes the code a lot more verbose (need to insert a bunch more calls to `walkMostAST`)

src/typescript-service.ts
beyang: Is there a way to test for this case in CI?

src/typescript-service.ts
beyang: If I omit `syncConfiguration`, I get no results back. To test this, remove this line and run `npm test`.

src/typescript-service.ts
beyang: This is meant to mean "effectively unlimited" and is only used once, so I inlined it instead of giving it a name.

src/typescript-service.ts
beyang: Changed to use relative URI in the connection handler

beyang:
@alexsaveliev PTAL

src/typescript-service.ts
alexsaveliev: > If I omit syncConfiguration, I get no results back. To test this, remove this line and run npm test.

Addressed in @327a735e16da2782439d717160817f966bb1ffdb
You should use our `TypescriptService::getSourceFile` instead of `Program::getSourceFile`, this way we'll put missing files to TS program and there will be no need to sync the whole configuration

alexsaveliev:
LGTM now, thanks! I have added some minor changes

PR #37, opened 2016-12-02
Minor refactor

beyang:


src/typescript-service.ts
```diff
@@ -198,7 +198,7 @@ export default class TypeScriptService {
 						const ret = [];
 						if (defs) {
 							for (let def of defs) {
-								const sourceFile = configuration.program.getSourceFile(def.fileName);
+								const sourceFile = this.getSourceFile(configuration, def.fileName);
```
alexsaveliev: This one may be redundant because if there was no source file known to TS there will be no definition produced. Does not do any harm on other hand :)

PR #38, opened 2016-12-06
Factor out LanguageHandler interface

beyang:
This PR does not change any behavior. It removes the `Connection` type, because it was an unnecessary layer between `IConnection` and `TypeScriptService` and introduces a new interface, `LanguageHandler`, that can be implemented by other `TypeScriptService`-like things (e.g., a version of `TypeScriptService` that handles fetching external dependencies).

### DIFF
--- a/src/connection.ts
+++ b/src/connection.ts
@@ -7,6 +7,7 @@ import {
 	InitializeResult,
 	TextDocuments,
 	TextDocumentPositionParams,
+	TextDocumentSyncKind,
 	Definition,
 	ReferenceParams,
 	Location,
@@ -30,7 +31,6 @@ import TypeScriptService from './typescript-service';
 
 import * as rt from './request-type';
 
-
 /**
  * Connection handles incoming requests and sends responses over the
  * JSONRPC connection. There is one Connection instance created for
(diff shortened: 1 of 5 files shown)


src/connection.ts
```diff
@@ -26,227 +25,147 @@ import * as ts from 'typescript';
 import * as types from 'vscode-languageserver-types';
 
 import * as util from './util';
-import TypeScriptService from './typescript-service';
+import { LanguageHandler, TypeScriptService } from './typescript-service';
 
 import * as rt from './request-type';
 
+export default function NewConnection(input: any, output: any, strict: boolean, handler: LanguageHandler): IConnection {
```
alexsaveliev: I smell a ~~rat~~ gopher here :) Why not class?

src/typescript-service.ts
```diff
@@ -14,7 +36,37 @@ import * as rt from './request-type';
...
 
-export default class TypeScriptService {
+
+/**
+ * LanguageHandler handles LSP requests. It includes a handler method
+ * for each LSP method that this language server supports. Each
+ * handler method should be registered to the corresponding
+ * registration method on IConnection.
+ */
+export interface LanguageHandler {
```
alexsaveliev: I'd suggest to move interface to a separate `.ts` file

beyang:
Addressed all your comments

PR #39, opened 2016-12-09
Refactoring to enable language server to be used by buildserver

beyang:
This should involve no change in behavior, but it moves references to `IConnection` out of the language handler level. The only way that the language handler was using the connection was to issue LSP-VFS queries, which the build server will want to intercept. This change creates the remote FS outside of the language handler and then passes the remote FS to the language handler in the `initialize` method.

### DIFF
--- a/src/project-manager.ts
+++ b/src/project-manager.ts
@@ -68,15 +68,15 @@ export class ProjectManager {
 	/**
 	 * Ensures that all files are added (and parsed) to the project to which fileName belongs. 
 	 */
-	syncConfigurationFor(fileName: string, connection: IConnection) {
-		return this.syncConfiguration(this.getConfiguration(fileName), connection);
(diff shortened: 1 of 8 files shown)


src/project-manager.ts
```diff
@@ -34,19 +32,19 @@ export class ProjectManager {
...
 		this.versions = new Map<string, number>();
 		this.fetched = new Set<string>();
+		this.remoteFs = remoteFs;
 
-		if (strict) {
-			this.remoteFs = new FileSystem.RemoteFileSystem(connection)
-		} else {
-			this.remoteFs = new FileSystem.LocalFileSystem(root)
-		}
+		// if (strict) {
```
alexsaveliev: Why leave commented code?

beyang:
Rebased.

PR #44, opened 2016-12-27
Improve perf of hover and definition

beyang:
This improves perf of hover and definition operations by giving a given LSP connection access to two workers (each of which has one CPU).

### DIFF
--- a/src/server.ts
+++ b/src/server.ts
@@ -1,0 +1,108 @@
+/// <reference path="../node_modules/vscode/thenable.d.ts" />
+
+import {
+	IConnection,
+} from 'vscode-languageserver';
+
+import * as net from 'net';
+import * as cluster from 'cluster';
+
+import { newConnection, registerLanguageHandler } from './connection';
+import { registerMasterHandler } from './master-connection';
+import { LanguageHandler } from './lang-handler';
+import * as fs from './fs';
+
+const workersReady = new Map<string, Promise<void>>();
+
+function randomNWorkers(n: number): string[] {
+	const unselected = Array.from(workersReady.keys());
+	let numUnselected = unselected.length;
+	const selected: string[] = [];
+	for (let i = 0; i < n; i++) {
+		const s = Math.floor(Math.random() * numUnselected);
+		selected.push(unselected[s]);
+		const a = unselected[numUnselected - 1], b = unselected[s];
+		unselected[numUnselected - 1] = b, unselected[s] = a;
+		numUnselected--;
+	}
+	return selected;
+}
+
+async function rewriteConsole() {
+	const consoleErr = console.error;
+	console.error = function () {
+		if (cluster.isMaster) {
+			consoleErr(`[mstr]`, ...arguments);
+		} else {
+			consoleErr(`[wkr${cluster.worker.id}]`, ...arguments);
+		}
+	}
+}
+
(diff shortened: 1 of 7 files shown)


src/server.ts
```diff
@@ -0,0 +1,108 @@
...
+		cluster.on('exit', (worker, code, signal) => {
+			const reason = code === null ? signal : code;
+			console.error(`worker ${worker.process.pid} exit (${reason})`);
+		});
+
+		var server = net.createServer(async (socket) => {
+			const connection = newConnection(socket, socket);
+
+			// Create connections to two worker servers
+			const workerIds = randomNWorkers(2);
```
alexsaveliev: My understanding is that we are using 2 workers from N where N is a number of CPUs. 
- Why not all of them?
- If we want to stick to 2 why create N workers to peek random 2 from them?
- If we want to stick to 2 we should ensure we have at least two CPUs so two workers can be started
- If we want to stick to 2 there is no sense to use cluster size command line parameter

beyang: Previously, we'd have N workers, but a given connection would be tied to a single worker. As a result, LSP operations for a given workspace were all single-threaded. This means that one long-running operation (e.g., a `workspace/symbol` invocation) could block other operations (e.g., `hover` or `definition`) from completing.

With the new code, each connection has 2 workers, one which only receives calls to quick endpoints like `definition` and `hover`. We could increase the number of workers per connection, but this would eat up more CPU per connection (which could cause slowness if the LSP has multiple workspaces open) while not really yielding any additional performance gain.

src/typescript-service.ts
```diff
@@ -1145,69 +1137,41 @@ export class TypeScriptService implements LanguageHandler {
...
+					const items = config.getService().getNavigateToItems(query, undefined, undefined, true);
+					for (const item of items) {
+						symbols.push(this.transformNavItem(this.root, config.getProgram(), item));
+					}
+				} else {
+					Array.prototype.push.apply(symbols, this.getNavigationTreeItems(config));
 				}
-				this.collectWorkspaceSymbols(query, limit, configurations, index + 1, items, callback);
-			};
+			})
```
beyang: No semicolon allowed here (since the expression is inside `Promise.all(...)`)

src/typescript-service.ts
```diff
@@ -186,32 +186,24 @@ export class TypeScriptService implements LanguageHandler {
...
+			return this.emptyQueryWorkspaceSymbols;
+		}
+		const configs = this.projectManager.getConfigurations();
+		const itemsPromise = this.collectWorkspaceSymbols(query, configs);
+		if (!query) {
+			this.emptyQueryWorkspaceSymbols = itemsPromise;
+		}
+		let items = await itemsPromise
+		let sortedItems = items.sort((a, b) => a.name.toLocaleLowerCase().localeCompare(b.name.toLocaleLowerCase()));
+		let limitedItems = sortedItems.slice(0, limit);
```
beyang: There would be a performance benefit to doing so, but it is minor. The previous limit was 100, which meant as soon as the user entered a query for which there were fewer than 100 results (basically every query other than the empty one), we'd have to do as much work as collecting all items. So the performance benefit would really only be seen for the initial query. I decided to remove this behavior, as it simplifies the code.

PR #62, opened 2017-01-18
getPackages endpoint, workspace/symbol symbol query extension

beyang:


alexsaveliev:
Code LGTM but I have few remarks

- should we include dependencies that do not have `name: version range` form (`user/repo`, `url`, `local file`)?
- I'd rather try to resolve version range to version like npm/yarn does, if possible
- Also, would be great (if we will be able to resolve exact version) to map version to Git hash (see `gitHead`)

felixfbecker:
To give my input on this, in PHP I read the lock file to get the _resolved_ version instead of the range.

beyang:
Merged, will address your comments later.

Issue #58, opened 2017-01-16
Don't commit compiled JS


felixfbecker:
General pattern for TS libraries is to commit only .ts source and only publish .js, .d.ts and .js.map to npm. We should do it that way to prevent the risk of compilation results and source getting out of sync. In Sourcegraph we install from npm, we have control over the publish process so we can always do pre-releases and locally we can use `npm link`, so there seems to be no need for having the compiled files on GH, which is kind of like committing binaries.

alexsaveliev:
@felixfbecker Yesterday I've said that there are no places when somebody uses GitHub-style dependency left. Today we got one :smile: , see:
https://github.com/sourcegraph/sourcegraph/pull/3104/files#diff-32dd941b31b9224e0da05d34b31c9282

felixfbecker:
but why? if this is supposed to get merged, it should use a version reference. If it is used in local development, `npm link` should be used

PR #66, opened 2017-01-20
Excluding compiled JS files from project

alexsaveliev:
- Addresses #58
- Please note that there is no more way to reference this package as `user/repo` or `clone-URL` style, only through registry or `npm link`
- Made `node bin/build` work standalone by providing required information from `package.json`


felixfbecker:
You should add a .npmignore that is the inverse, eg excludes `src`. Otherwise npm will use the .gitignore as a .npmignore and TS can also get confused with imports.

alexsaveliev:
> You should add a .npmignore that is the inverse, eg excludes src. Otherwise npm will use the .gitignore as a .npmignore and TS can also get confused with imports.

Nope, `npm pack` or `npm publish` produces tarball based on `files` section of `package.json`.
There is an issue with `yarn pack` because yarn does not process folders in the `files` section but this is a different story :smile: 

> TS can also get confused with imports.

Sorry, I do not follow you here.. Could you please give a sample?

PR #67, opened 2017-01-20
Empty files are valid files too

alexsaveliev:
- when FS read empty file it's OK, we should compare against `undefined` to determine if there is no such a file instead of casting content to boolean. Addresses #64
- made `LanguageServiceHost::log` and `LanguageServiceHost::trace` implementations empty, otherwise `completions` request may produce lot of useless output


src/project-manager.ts
```diff
@@ -573,11 +573,11 @@ export class InMemoryLanguageServiceHost implements ts.LanguageServiceHost {
 
 	getScriptSnapshot(fileName: string): ts.IScriptSnapshot {
 		let entry = this.fs.readFile(fileName);
```
felixfbecker: `fs.readFile` should be typed to return `string | undefined`

PR #68, opened 2017-01-25
e2e tests (WIP)

beyang:
- adding e2e tests that clone specific repository/revision and then testing LSP protocol using cloned repository as workspace root
- fixed flaky packages test (packages must be sorted)
- added in-memory file system implementation
- added ability to use specific resolver to construct path from components. The reason is that on Windows in VFS mode there may be a request to read content of "/" directory backed by the local filesystem which causes `path.resolve(root, component)` to produce `/` AKA `C:\`
- added possibility to run e2e tests only or exclude them by using npm scripts `test-e2e` and `test-lsp`

### DIFF
--- a/bin/test
+++ b/bin/test
@@ -6,8 +6,13 @@ let path = require('path');
 
 process.chdir(path.dirname(__dirname));
 
+let filter = '';
+if (process.argv.length > 2) {
+	filter = '-g ' + process.argv[2];
+}
+
 try {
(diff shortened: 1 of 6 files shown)


felixfbecker:
@beyang can we close this? I think we would want e2e tests like we have for Go, the implementation has countless merge conflicts and is WIP. I also think cloning is not a good idea because it depends on external services and can make tests flaky.

alexsaveliev:
@felixfbecker as an author, I am ok to close it :)

PR #75, opened 2017-02-22
Hand off requests to child processes by effort

felixfbecker:
For every TCP worker (provided by node's cluster module) two child processes are spawned, one for long-running requests like references, the other for short-running requests like definition. The spawning is provided by a passed function so the build server can spawn a build server instead (in the future I would like to refactor this a lot to be more object-oriented).

The child processes talk LSP over STDIO, the TCP worker handles proxying messages to/from the socket.

Currently the processes are spawned on an incoming TCP connection, it would be faster to keep them warm, but I need to drop the exit notifications for this and check if the LS supports re-initialization for this.

### DIFF
--- a/src/connection.ts
+++ b/src/connection.ts
@@ -34,7 +34,7 @@ export interface TraceOptions {
 	logfile?: string;
 }
 
-export function newConnection(input: any, output: any, trace: TraceOptions): IConnection {
+export function newConnection(input: NodeJS.ReadableStream, output: NodeJS.WritableStream, trace: TraceOptions): IConnection {
 
 	const reader = new StreamMessageReader(input);
 
--- a/src/connection.ts
+++ b/src/connection.ts
@@ -64,29 +64,12 @@ export function newConnection(input: NodeJS.ReadableStream, output: NodeJS.Writa
 	}
 
 	const connection = createConnection(reader, writer);
+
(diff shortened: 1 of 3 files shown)


beyang:
Have you evaluated how much faster this makes things in practice across a set of repositories? I'm curious to see if the extra process startup time (100ms? 1s?) impacts initialization time.

felixfbecker:
The spawn call takes 1-4ms

felixfbecker:
Hit some issues with child processes not exiting properly which I need to resolve before merging

beyang:
Sounds good, LGTM otherwise.

src/connection.ts
```diff
@@ -64,30 +64,13 @@ export function newConnection(input: any, output: any, trace: TraceOptions): ICo
...
-			}
-			closed = true;
-		}
-	}
-
-	// We attach one notification handler on `exit` here to handle the
-	// teardown of the connection.  If other handlers want to do
-	// something on connection destruction, they should register a
-	// handler on `shutdown`.
-	connection.onNotification(rt.ExitRequest.type, close);
```
felixfbecker: this is now handled in `server.ts` so we close the socket correctly (it's `socket.end()`, not `socket.close()`) and we don't override the exit handler (as vscode-languageserver/json-rpc only allows a single handler per method). Previously because of this code here the STDIO version would not exit on an exit notification, which the child processes need to do.

PR #79, opened 2017-02-24
Restart workers that died

felixfbecker:
Does not implement any restart logic for the child processes.

### DIFF
--- a/src/server.ts
+++ b/src/server.ts
@@ -37,12 +37,12 @@ export async function serve(options: ServeOptions, createLangHandler: () => Lang
 	rewriteConsole();
 
 	if (cluster.isMaster) {
-		console.error(`Master node process spawning ${options.clusterSize} workers`);
+		console.error(`Master (PID ${process.pid}) spawning ${options.clusterSize} workers`);
 		cluster.on('online', worker => {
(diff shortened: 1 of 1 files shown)


src/server.ts
```diff
@@ -37,17 +37,17 @@ export async function serve(options: ServeOptions, createLangHandler: () => Lang
...
-
+		console.error(`Master node process spawning ${options.clusterSize} workers`);
+		cluster.on('online', worker => {
+			console.error(`Worker ${worker.id} online`);
+		});
 		cluster.on('exit', (worker, code, signal) => {
-			const reason = code === null ? signal : code;
-			console.error(`worker ${worker.process.pid} exit (${reason})`);
+			console.error(`Worker ${worker.id} exited from signal ${signal} with code ${code}, restarting`);
+			cluster.fork();
```
beyang: Would this cause issues with legitimate killing of workers? E.g., in dev you kill the master and the workers respawn indefinitely.

beyang: oh never mind, I misread this.

felixfbecker: This will not happen if the master is killed, but in any other case the worker is killed, yes

PR #82, opened 2017-03-03
Big refactor including move from legacy vfs to workspace/files extension

### LINKED ISSUE #77: Switch from legacy fs extension to new files extension
The old `fs/readFile` extension should be replaced by the new `workspace/files`, `textDocument/content` extension.

https://github.com/sourcegraph/language-server-protocol/blob/master/extension-files.md

felixfbecker:
Closes #77

- Switches from the legacy vfs extension to workspace/files
- Refactors code to use promises more and less callbacks
- Simplifies tests
  - Use async functions instead of callbacks
  - Call TypeScriptService directly
  - Remove test utils that just composed callbacks and used `assert.deepEqual()`
- Remove bin/ scripts. They can all be expressed in a one-line package.json script.
- Remove tsfmt. It is too buggy for my experience and TSLint does a much better job for style checking
- Fix some bugs I discovered with hovers
- Remove `process.on('uncaughtException')`
- Improves every bit of code I came across.


src/project-manager.ts
```diff
@@ -413,56 +324,30 @@ export class ProjectManager {
...
+	 *
+	 * @param files File paths
+	 */
+	async ensureFiles(files: string[]): Promise<void> {
+
+		// Only fetch files that are not already fetched
+		files = files.filter(file => !this.fetched.has(file));
+
+		await bluebird.map(files, async path => {
+			const content = await this.remoteFs.getTextDocumentContent(util.path2uri('', path))
```
eseliger: `;`

package.json
```diff
@@ -25,26 +25,36 @@
...
+    "watch": "tsc -w"
   },
   "dependencies": {
-    "@types/async": "^2.0.32",
+    "@types/async": "^2.0.38",
+    "@types/bluebird": "^3.0.37",
+    "@types/chai-as-promised": "^0.0.29",
+    "@types/glob": "^5.0.30",
+    "@types/mz": "^0.0.30",
     "@types/node": "^6.0.46",
```
eseliger: Why aren't those typings `devDependencies`?

src/test/typescript-service-helpers.ts
```diff
@@ -0,0 +1,1525 @@
...
+			});
+		});
+		it('produces completions for referenced symbols', <any>async function (this: TestContext) {
+			const result = await this.service.getCompletions({
+				textDocument: {
+					uri: 'file:///uses-reference.ts'
+				},
+				position: {
+					line: 1,
+					character: 13
```
beyang: I'm not going to verify the expected test data, but I assume you've verified that the expected data hasn't substantively changed?

src/typescript-service.ts
```diff
@@ -51,37 +52,37 @@ export class TypeScriptService implements LanguageHandler {
...
-						},
-						xpackagesProvider: true,
-					}
-				})
+		if (params.rootPath) {
+			this.root = util.uri2path(params.rootPath);
+			this.strict = strict;
+			this.projectManager = new pm.ProjectManager(this.root, remoteFs, strict, this.traceModuleResolution);
+			// Pre-fetch files in the background
+			// TODO why does ensureAllFiles() fetch less files than ensureFilesForWorkspaceSymbol()?
```
beyang: remove TODO?

src/typescript-service.ts
```diff
@@ -51,37 +52,37 @@ export class TypeScriptService implements LanguageHandler {
 		this.traceModuleResolution = traceModuleResolution || false;
 	}
 
+	// TODO pass remoteFs and strict to constructor
```
beyang: remove TODO?

src/fs.ts
```diff
@@ -1,86 +1,90 @@
...
-	readDir(path: string, callback: (err: Error | null, result?: FileInfo[]) => void) {
-		this.connection.sendRequest(ReadDirRequest.type, path).then((f: FileInfo[]) => {
-			return callback(null, f)
-		}, callback)
+	/**
+	 * The files request is sent from the server to the client to request a list of all files in the workspace or inside the directory of the base parameter, if given.
+	 * A language server can use the result to index files by filtering and doing a content request for each text document of interest.
+	 */
+	getWorkspaceFiles(base?: string): Promise<string[]> {
+		// TODO cache this at a different layer and invalidate it properly
```
beyang: Why? At what layer would you cache it?

src/project-manager.ts
```diff
@@ -148,6 +142,7 @@ export class ProjectManager {
 		}
 	}
 
+	// TODO replace this with something like a WorkspaceTree class
```
beyang: Not sure if it is worth the trouble? What advantage would it bring over the current implementation?

beyang:
A few minor comments, mostly on the TODO statements. Otherwise LGTM.

src/test/typescript-service-helpers.ts
felixfbecker: It's the same test file adapted, didn't change the expected values (except for legitimate cases where eg `range` was missing or the return value of `getDefinition` has to be an array)

src/typescript-service.ts
felixfbecker: Do you consider it bad practice to add TODOs to the code?

src/fs.ts
felixfbecker: What I don't like here is that we keep an array of the files here in memory, but `LocalFileSystem` essentially does too (with the difference that the former is never updated). I was thinking about a `WorkspaceTree` class that holds this state and optionally the content of files. Maybe also remove the dummy placeholder contents.

src/typescript-service.ts
beyang: Nope, I misread this as meaning "pass remoteFs and strict to the ProjectManager constructor" and saw that they were already passed and the TODO was satisfied.

felixfbecker: I meant to not pollute the `initialize` arguments because they should be used for other stuff (cancellation token etc) and should be consistent with other method handlers

src/typescript-service.ts
felixfbecker: I would like to clean this up eventually as I think the current situation is either a bug or simply not documented enough / confusing behaviour.

PR #97, opened 2017-03-13
Code improvements

alexsaveliev:
- documented the code
- some fields/arguments were renamed to be self explanatory
- replaced `any` with typed objects
- fixed tests on Windows 
- restored Appveyor integration
- addresses #98

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,9 +7,9 @@ cache:
   yarn: true
 
 script:
-  - yarn run lint
-  - yarn run build
-  - yarn run cover
+  - npm run lint
+  - npm run build
+  - npm run cover
 
 after_success:
   # Upload coverage to codecov
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -16,8 +16,7 @@ test_script:
   - npm --version
   # run tests
   - yarn run lint
-  - yarn run build
   - yarn run cover
 
-# Don't actually build.
-build: off
+build_script:
+  - yarn run build
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -15,8 +15,9 @@ test_script:
   - node --version
   - npm --version
   # run tests
-  - yarn test
-  - yarn fmt-check
+  - yarn run lint
+  - yarn run build
+  - yarn run cover
 
 # Don't actually build.
 build: off
--- a/src/project-manager.ts
+++ b/src/project-manager.ts
@@ -504,12 +504,12 @@ export class ProjectManager implements Disposable {
 	 */
 	refreshConfigurations() {
 		const rootdirs = new Set<string>();
-		for (const k of Object.keys(this.localFs.entries)) {
+		this.localFs.entries.forEach((v, k) => {
 			if (!/(^|\/)[tj]sconfig\.json$/.test(k)) {
-				continue;
+				return;
 			}
 			if (/(^|\/)node_modules\//.test(k)) {
-				continue;
+				return;
 			}
 			let dir = path_.posix.dirname(k);
 			if (dir === '.') {
@@ -517,7 +517,7 @@ export class ProjectManager implements Disposable {
 			}
 			this.configs.set(dir, new ProjectConfiguration(this.localFs, path_.posix.join('/', dir), this.versions, k, undefined, this.traceModuleResolution));
 			rootdirs.add(dir);
-		}
+		});
 		if (!rootdirs.has('')) {
 			// collecting all the files in workspace by making fake configuration object
 			this.configs.set('', new ProjectConfiguration(this.localFs, '/', this.versions, '', {
@@ -589,7 +589,7 @@ export class InMemoryLanguageServiceHost implements ts.LanguageServiceHost {
 		this.files = [];
 		// adding library files from the local file system
 		getTypeScriptLibraries().forEach((content, name) => {
-			this.fs.entries[name] = content;
+			this.fs.entries.set(name, content);
 		});
 	}
 
@@ -678,7 +678,7 @@ const localFSPlaceholder = 'var dummy_0ff1bd;';
  */
 export interface FileSystemNode {
 	file: boolean;
-	children: {[directory: string]: FileSystemNode};
+	children: Map<string, FileSystemNode>;
 };
 
 /**
@@ -689,12 +689,12 @@ export class InMemoryFileSystem implements ts.ParseConfigHost, ts.ModuleResoluti
 	/**
 	 * Map (filepath -> string content) of files fetched from the remote file system. Paths are relative to `this.path`
 	 */
-	entries: { [filePath: string]: string };
+	entries: Map<string, string>;
 
 	/**
 	 * Map (filepath -> string content) of temporary files made while user modifies local file(s).  Paths are relative to `this.path`
 	 */
-	overlay: { [filePath: string]: string };
+	overlay: Map<string, string>;
 
 	/**
 	 * Should we take into account register when performing a file name match or not. On Windows when using local file system, file names are case-insensitive
@@ -713,9 +713,9 @@ export class InMemoryFileSystem implements ts.ParseConfigHost, ts.ModuleResoluti
 
 	constructor(path: string) {
 		this.path = path;
-		this.entries = {};
-		this.overlay = {};
-		this.rootNode = {file: false, children: {}};
+		this.entries = new Map<string, string>();
+		this.overlay = new Map<string, string>();
+		this.rootNode = {file: false, children: new Map<string, FileSystemNode>()};
 	}
 
 	/**
@@ -724,16 +724,17 @@ export class InMemoryFileSystem implements ts.ParseConfigHost, ts.ModuleResoluti
 	 * @param content file content
 	 */
 	addFile(path: string, content: string) {
-		this.entries[path] = content;
+		this.entries.set(path, content);
 		let node = this.rootNode;
 		path.split('/').forEach((component, i, components) => {
-			const n = node.children[component];
+			const n = node.children.get(component);
 			if (!n) {
 				if (i < components.length - 1) {
-					node.children[component] = {file: false, children: {}};
-					node = node.children[component];
+					const n = {file: false, children: new Map<string, FileSystemNode>()};
+					node.children.set(component, n);
+					node = n;
 				} else {
-					node.children[component] = {file: true, children: {}};
+					node.children.set(component, {file: true, children: new Map<string, FileSystemNode>()});
 				}
 			} else {
 				node = n;
@@ -754,31 +755,30 @@ export class InMemoryFileSystem implements ts.ParseConfigHost, ts.ModuleResoluti
 	 * @return file's content in the following order (overlay then cache) if any
 	 */
 	readFile(path: string): string | undefined {
-		let content = this.overlay[path];
+		let content = this.overlay.get(path);
 		if (content !== undefined) {
 			return content;
 		}
 
 		const rel = path_.posix.relative('/', path);
-
-		content = this.overlay[rel];
+		content = this.overlay.get(rel);
 		if (content !== undefined) {
 			return content;
 		}
 
-		content = this.entries[path];
+		content = this.entries.get(path);
 		if (content !== undefined) {
 			return content;
 		}
 
-		return this.entries[rel];
+		return this.entries.get(rel);
 	}
 
 	/**
 	 * Invalidates temporary content denoted by the given path
 	 */
 	didClose(path: string) {
-		delete this.overlay[path];
+		this.overlay.delete(path);
 	}
 
 	/**
@@ -792,7 +792,7 @@ export class InMemoryFileSystem implements ts.ParseConfigHost, ts.ModuleResoluti
 	 * Updates temporary content denoted by the given path
 	 */
 	didChange(path: string, text: string) {
-		this.overlay[path] = text;
+		this.overlay.set(path, text);
 	}
 
 	/**
@@ -817,15 +817,15 @@ export class InMemoryFileSystem implements ts.ParseConfigHost, ts.ModuleResoluti
 		const components = path.split('/').filter(c => c);
 		if (components.length !== 1 || components[0]) {
 			for (const component of components) {
-				const n = node.children[component];
+				const n = node.children.get(component);
 				if (!n) {
 					return ret;
 				}
 				node = n;
 			}
 		}
-		Object.keys(node.children).forEach(name => {
-			if (node.children[name].file) {
+		node.children.forEach((value, name) => {
+			if (value.file) {
 				ret.files.push(name);
 			} else {
 				ret.directories.push(name);
--- a/src/project-manager.ts
+++ b/src/project-manager.ts
@@ -34,7 +34,8 @@ export class ProjectManager implements Disposable {
 
 	/**
 	 * Workspace subtree (folder) -> JS/TS configuration mapping.
-	 * Configuration settings for a source file A are located in the closest parent folder of A
+	 * Configuration settings for a source file A are located in the closest parent folder of A.
+	 * Map keys are relative (to workspace root) paths
 	 */
 	private configs: Map<string, ProjectConfiguration>;
 
@@ -56,7 +57,7 @@ export class ProjectManager implements Disposable {
 	private localFs: InMemoryFileSystem;
 
 	/**
-	 * URI -> version map. Every time file content is about to change or changed (didChange/didOpen/...), we are incrementing it's version
+	 * Relative file path -> version map. Every time file content is about to change or changed (didChange/didOpen/...), we are incrementing it's version
 	 * signalling that file is changed and file's user must invalidate cached and requery file content
 	 */
 	private versions: Map<string, number>;
@@ -73,6 +74,8 @@ export class ProjectManager implements Disposable {
 	 * won't fetch it again. This should be cleared if remoteFs files
 	 * have been modified in some way, but does not need to be cleared
 	 * if remoteFs files have only been added.
+	 *
+	 * Set elements are absolute file paths
 	 */
 	private fetched: Set<string>;
 
@@ -135,6 +138,7 @@ export class ProjectManager implements Disposable {
 	}
 
 	/**
+	 * @param filePath file path (both absolute or relative file paths are accepted)
 	 * @return true if there is a fetched file with a given path
 	 */
 	hasFile(filePath: string) {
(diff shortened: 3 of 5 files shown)


src/project-manager.ts
```diff
@@ -528,11 +679,30 @@ const localFSPlaceholder = 'var dummy_0ff1bd;';
  */
 export class InMemoryFileSystem implements ts.ParseConfigHost, ts.ModuleResolutionHost {
 
+	/**
+	 * It's actually a map (filepath -> string content) of files fetched from the remote file system
```
felixfbecker: So it is not a tree structure?

Is this correct?
```ts
entries: { [filePath: string]: string };
```

src/project-manager.ts
```diff
@@ -528,11 +679,30 @@ const localFSPlaceholder = 'var dummy_0ff1bd;';
...
+
+	/**
+	 * Should we take into account register when performing a file name match or not. On Windows when using local file system, file names are case-insensitive 
+	 */
 	useCaseSensitiveFileNames: boolean;
+
+	/**
+	 * Root URI
+	 */
 	path: string;
```
felixfbecker: Is this part of the file paths in `entries`?

src/project-manager.ts
```diff
@@ -528,11 +679,30 @@ const localFSPlaceholder = 'var dummy_0ff1bd;';
...
 
+	/**
+	 * It's a tree node where properties point to either a file (typeof node[p] == "string") or to a sub-tree.
+	 * We are using it when TS service scans for source files that making some project
+	 */
 	rootNode: any;
```
felixfbecker: Is this correct?
```ts
type Node = string | { [directory: string]: Node }

rootNode: Node;
```

Does a string value contain the file content? If yes, why do we have both a tree and the `entries` map?

In `addFile` there is sometimes an asterisk used for the key, what meaning does this have?

src/project-manager.ts
```diff
@@ -731,33 +964,51 @@ export class ProjectConfiguration {
...
 	syncProgram(): void {
 		this.program = this.getService().getProgram();
 	}
 
 	private initialized?: Promise<void>;
 
+	/**
+	 * Initializes (sub)project by parsing configuration and making proper internal objects
+	 */
 	private init(): Promise<void> {
```
felixfbecker: Is there a reason this is asynchronous? It looks like it doesn't do anything async, so couldn't this simply be the constructor of the object?

src/project-manager.ts
```diff
@@ -26,15 +26,44 @@ export class ProjectManager implements Disposable {
 	 * Cancellations to do when the object is disposed
 	 */
 	private cancellationSources = new Set<CancellationTokenSource>();
+
+	/**
+	 * Root URI (as passed to `initialize` request)
```
alexsaveliev: You are right, this is root path. Renamed field and parameter name

src/project-manager.ts
```diff
@@ -71,23 +120,31 @@ export class ProjectManager implements Disposable {
...
+	 */
 	getFs(): InMemoryFileSystem {
 		return this.localFs;
 	}
 
 	/**
-	 * @return true if there is a file with a given name
+	 * @return true if there is a fetched file with a given name
 	 */
 	hasFile(name: string) {
```
alexsaveliev: yes, renamed

src/project-manager.ts
alexsaveliev: Yes, updated

src/project-manager.ts
alexsaveliev: Nope, paths in `entries` are relative to `path`. Updated

src/project-manager.ts
alexsaveliev: replaced `any` with 
```export type FileSystemNode = { 
	file: boolean;
	children: {[directory: string]: FileSystemNode} 
};
```

> Does a string value contain the file content? If yes, why do we have both a tree and the entries map?

Removed string content from a tree, left structure only

> In addFile there is sometimes an asterisk used for the key, what meaning does this have?

Just string / not string, obsolete now

src/project-manager.ts
alexsaveliev: Yes, it can be replaced with a constructor in you want

appveyor.yml
```diff
@@ -15,8 +15,9 @@ test_script:
   - node --version
   - npm --version
   # run tests
-  - yarn test
-  - yarn fmt-check
+  - yarn run lint
+  - yarn run build
```
felixfbecker: Shouldn't this be the `build` phase that is currently deactivated below?

src/project-manager.ts
```diff
@@ -26,15 +26,44 @@ export class ProjectManager implements Disposable {
...
 	private cancellationSources = new Set<CancellationTokenSource>();
-	private root: string;
+
+	/**
+	 * Root path (as passed to `initialize` request)
+	 */
+	private rootPath: string;
+
+	/**
+	 * Workspace subtree (folder) -> JS/TS configuration mapping.
```
felixfbecker: are the folders relative to `rootPath`? Are they paths or URIs?

src/project-manager.ts
```diff
@@ -47,14 +76,34 @@ export class ProjectManager implements Disposable {
 	 */
 	private fetched: Set<string>;
```
felixfbecker: are these file paths or URIs? relative or absolute?

src/project-manager.ts
```diff
@@ -71,23 +120,31 @@ export class ProjectManager implements Disposable {
...
 		return this.localFs;
 	}
 
 	/**
-	 * @return true if there is a file with a given name
+	 * @return true if there is a fetched file with a given path
 	 */
-	hasFile(name: string) {
-		return this.localFs.fileExists(name);
+	hasFile(filePath: string) {
```
felixfbecker: Is this path relative to root path?

src/project-manager.ts
```diff
@@ -238,14 +321,27 @@ export class ProjectManager implements Disposable {
...
+	/**
+	 * Recursively collects file(s) dependencies up to given level.
+	 * Dependencies are extracted by TS compiler from import and reference statements
+	 *
+	 * Dependencies include:
+	 * - all the configuration files
+	 * - files referenced by the given file
+	 * - files included by the given file
+	 *
+	 * @param fileNames files to process
```
felixfbecker: I assume these are file paths - are they relative to root path?

src/project-manager.ts
```diff
@@ -284,12 +380,13 @@ export class ProjectManager implements Disposable {
 	}
 
 	/**
+	 * @param filePath source file path
```
felixfbecker: relative to root path?

src/project-manager.ts
```diff
@@ -306,10 +403,20 @@ export class ProjectManager implements Disposable {
 		throw new Error('unreachable');
 	}
 
+	/**
+	 * Called when file was opened by client. Current implementation
+	 * does not differenciates open and change events
+	 * @param fileName path to a file
```
felixfbecker: The method looks like the LSP method, but the LSP method takes a URI. You sure this is a path? If it's a path, is it relative to the root path?

(same for all `did*` methods)

src/project-manager.ts
```diff
@@ -321,6 +428,11 @@ export class ProjectManager implements Disposable {
...
 	}
 
+	/**
+	 * Called when file was changed by client. Current implementation invalidates compiled version
+	 * @param fileName path to a file
+	 * @param text file's content
+	 */
 	didChange(fileName: string, text: string) {
 		this.localFs.didChange(fileName, text);
 		let version = this.versions.get(fileName) || 0;
```
felixfbecker: `versions` is documented as a Map from URI to version, but `fileName` is documented as a file path

src/project-manager.ts
```diff
@@ -372,18 +488,19 @@ export class ProjectManager implements Disposable {
...
-				// There may be too many open files when working with local FS and trying
-				// to open them in parallel, so limit concurrent readFile calls to 100
-				// TODO only do this when working with localFs?
-				concurrency: 100
-			});
+					// There may be too many open files when working with local FS and trying
+					// to open them in parallel, so limit concurrent readFile calls to 100
+					// TODO only do this when working with localFs?
+					concurrency: 100
+				});
```
felixfbecker: formatting error here (I assume this was tsfmt, this is the reason why I consider tsfmt broken)

src/project-manager.ts
```diff
@@ -415,21 +532,51 @@ export class ProjectManager implements Disposable {
...
+ * It takes file content from local cache and provides it to TS compiler on demand
+ *
+ * @implements ts.LanguageServiceHost
  */
 export class InMemoryLanguageServiceHost implements ts.LanguageServiceHost {
 
 	complete: boolean;
 
+	/**
+	 * Root URI
```
felixfbecker: URI or path?

src/project-manager.ts
```diff
@@ -415,21 +532,51 @@ export class ProjectManager implements Disposable {
...
+	/**
+	 * List of files that project consist of (based on tsconfig includes/excludes and wildcards)
+	 */
 	expectedFiles: string[];
 
+	/**
+	 * Current list of files that were implicitly added to project
+	 * (every time when we need to extract data from a file that we havem't touched yet)
+	 */
 	private files: string[];
```
felixfbecker: paths or URIs? relative to root path?
Do these have to be ordered or could this be a `Set`?

src/project-manager.ts
```diff
@@ -415,21 +532,51 @@ export class ProjectManager implements Disposable {
...
+
+	/**
+	 * Local file cache where we looking for file content
+	 */
 	private fs: InMemoryFileSystem;
+
+	/**
+	 * List of files that project consist of (based on tsconfig includes/excludes and wildcards)
+	 */
 	expectedFiles: string[];
```
felixfbecker: file paths or URIs? relative to root path?

src/project-manager.ts
```diff
@@ -524,42 +674,85 @@ export class InMemoryLanguageServiceHost implements ts.LanguageServiceHost {
...
+
+/**
  * In-memory file system, can be served as a ParseConfigHost (thus allowing listing files that belong to project based on tsconfig.json options)
  */
 export class InMemoryFileSystem implements ts.ParseConfigHost, ts.ModuleResolutionHost {
 
-	entries: any;
-	overlay: any;
+	/**
+	 * Map (filepath -> string content) of files fetched from the remote file system. Paths are relative to `this.path`
```
felixfbecker: relative to root path?

src/project-manager.ts
```diff
@@ -524,42 +674,85 @@ export class InMemoryLanguageServiceHost implements ts.LanguageServiceHost {
...
+	 */
+	entries: { [filePath: string]: string };
+
+	/**
+	 * Map (filepath -> string content) of temporary files made while user modifies local file(s).  Paths are relative to `this.path`
```
felixfbecker: relative to root path?

src/project-manager.ts
```diff
@@ -524,42 +674,85 @@ export class InMemoryLanguageServiceHost implements ts.LanguageServiceHost {
...
+	 * File tree root
+	 */
+	rootNode: FileSystemNode;
 
 	constructor(path: string) {
 		this.path = path;
 		this.entries = {};
 		this.overlay = {};
-		this.rootNode = {};
+		this.rootNode = {file: false, children: {}};
```
felixfbecker: This will cause errors for folders or files that are named after a built in `Object` property, for example `hasOwnProperty`, `valueOf`, `toString`. `children` must be `Object.create(null)`

Can do this in a follow-up PR though.

src/project-manager.ts
```diff
@@ -524,42 +674,85 @@ export class InMemoryLanguageServiceHost implements ts.LanguageServiceHost {
...
 	}
 
+	/**
+	 * Adds file content to a local cache
+	 * @param path file path
```
felixfbecker: relative to root path?

src/project-manager.ts
```diff
@@ -524,42 +674,85 @@ export class InMemoryLanguageServiceHost implements ts.LanguageServiceHost {
...
+	/**
+	 * Tells if a file denoted by the given name exists in the local cache
+	 * @param path file path
+	 */
 	fileExists(path: string): boolean {
 		return this.readFile(path) !== undefined;
 	}
 
+	/**
+	 * @param path file path
```
felixfbecker: relative to root path?

src/project-manager.ts
```diff
@@ -680,13 +891,39 @@ export class ProjectConfiguration {
...
+	 */
 	private host?: InMemoryLanguageServiceHost;
 
+	/**
+	 * Local file cache
+	 */
 	private fs: InMemoryFileSystem;
+
+	/**
+	 * Path to configuration file (tsconfig.json/jsconfig.json)
```
felixfbecker: relative to root path?

src/project-manager.ts
```diff
@@ -680,13 +891,39 @@ export class ProjectConfiguration {
...
+	 */
 	private configFileName: string; // path to [tj]sconfig.json, if exists
+
+	/**
+	 * Configuration JSON object. May be used when there is no real configuration file to parse and use
+	 */
 	private configContent: any;
+
+	/**
+	 * Tracks source file -> version
```
felixfbecker: paths or URIs? relative to root?

src/project-manager.ts
```diff
@@ -680,13 +891,39 @@ export class ProjectConfiguration {
...
+	 */
 	private versions: Map<string, number>;
+
+	/**
+	 * Enables module resolution tracing (done by TS service)
+	 */
 	private traceModuleResolution: boolean;
+
+	/**
+	 * Root folder
```
felixfbecker: path or URI?

src/project-manager.ts
alexsaveliev: Relative paths. Updated documentation

src/project-manager.ts
alexsaveliev: Absolute file paths. Updated documentation

src/project-manager.ts
alexsaveliev: It accepts both relative and absolute ones. Updated documentation

src/project-manager.ts
alexsaveliev: Function accepts both relative and absolute paths. Updated documentation, renamed variables and arguments

src/project-manager.ts
alexsaveliev: yes, relative. Updated documentation

src/project-manager.ts
alexsaveliev: Yes, it's relative path. Updated documentation

src/project-manager.ts
alexsaveliev: My bad, updated `versions` doc

src/project-manager.ts
alexsaveliev: Reformatted. This was a result of VS Code auto format, not tsfmt

src/project-manager.ts
alexsaveliev: Path, sorry

src/project-manager.ts
felixfbecker: I think they call the same TS API

src/project-manager.ts
alexsaveliev: paths, relative. Updated documentation and renamed fields

src/project-manager.ts
alexsaveliev: relative, yes. Updated

src/project-manager.ts
alexsaveliev: relative, yes. Updated

src/project-manager.ts
alexsaveliev: Replaced with Map

src/project-manager.ts
alexsaveliev: relative, yes. Updated documentation

src/project-manager.ts
alexsaveliev: both styles are accepted. Mentioned that

src/project-manager.ts
alexsaveliev: Relative, updated documentation

src/project-manager.ts
alexsaveliev: path, relative. Updated documentation

src/project-manager.ts
alexsaveliev: path, relative to workspace hierarchy. Updated documentation and renamed field/parameter

appveyor.yml
alexsaveliev: Nope, Appveyor documentation https://www.appveyor.com/docs/lang/nodejs-iojs/ suggests doing it this way

felixfbecker: I was referring just to the `yarn run build` command.
The NodeJS documentation is meant for JavaScript, it doesn't compile any TypeScript.
It's nice to have AppVeyor report whether the compilation failed or tests failed

alexsaveliev: Moved `yarn build` to `build_script` section

.travis.yml
```diff
@@ -7,9 +7,9 @@ cache:
   yarn: true
 
 script:
-  - yarn run lint
-  - yarn run build
-  - yarn run cover
+  - npm run lint
```
beyang: why npm not yarn on Travis?

src/project-manager.ts
```diff
@@ -415,34 +536,66 @@ export class ProjectManager implements Disposable {
...
+	 * Each item is a relative file path
+	 */
+	expectedFilesPaths: string[];
+
+	/**
+	 * Current list of files that were implicitly added to project
+	 * (every time when we need to extract data from a file that we haven't touched yet).
+	 * Each item is a relative file path
+	 */
+	private filesPaths: string[];
```
beyang: s/filesPaths/filePaths/

src/project-manager.ts
```diff
@@ -415,34 +536,66 @@ export class ProjectManager implements Disposable {
...
+	 */
 	private fs: InMemoryFileSystem;
-	expectedFiles: string[];
 
-	private files: string[];
+	/**
+	 * List of files that project consist of (based on tsconfig includes/excludes and wildcards).
+	 * Each item is a relative file path
+	 */
+	expectedFilesPaths: string[];
```
beyang: s/expectedFilesPaths/expectedFilePaths/

.travis.yml
alexsaveliev: We experienced build errors with yarn on Travis today. Also, quoting @felixfbecker : 

> I don’t want to check in yarn.lock, because this is a package, not a project. if we check in yarn.lock it gives a false sense of security - the build can pass, but a user that `npm install`s the package could get a broken package

felixfbecker: Yarn seems to be bugged for some reason in Travis
https://travis-ci.org/sourcegraph/javascript-typescript-langserver/builds/210932923

If you know a fix, I have no problem with changing it back, but this shouldn't make any difference

Issue #76, opened 2017-02-23
Windows paths in references cause error in strict mode


felixfbecker:
TypeScript uses Windows-style paths in triple slash references:
```ts
/// <reference path="..\compiler\commandLineParser.ts"/>
```
https://sourcegraph.com/github.com/Microsoft/TypeScript@65da012527937a3074c62655d60ee08fee809f7f/-/blob/src/harness/rwcRunner.ts#L12:50

This causes an error when fetching the file in strict mode:

```
Unable to fetch content of /src/services/..\compiler\commandLineParser.ts { [Error:  /src/services/..\compiler\commandLineParser.ts: file does not exist]
  code: 0,
  message: ' /src/services/..\\compiler\\commandLineParser.ts: file does not exist',
  data: null }
```

cc @alexsaveliev

alexsaveliev:
claimed fixed

PR #101, opened 2017-03-15
Windows-style paths in triple slash references

alexsaveliev:
- handling Windows-style paths in triple slash references (addresses #76)

### DIFF
--- a/src/project-manager.ts
+++ b/src/project-manager.ts
@@ -104,7 +104,7 @@ export class ProjectManager implements Disposable {
 	 * @param traceModuleResolution allows to enable module resolution tracing (done by TS compiler)
 	 */
 	constructor(rootPath: string, remoteFs: FileSystem.FileSystem, strict: boolean, traceModuleResolution?: boolean) {
-		this.rootPath = util.normalizePath(rootPath);
+		this.rootPath = util.toUnixPath(rootPath);
 		this.configs = new Map<string, ProjectConfiguration>();
 		this.localFs = new InMemoryFileSystem(this.rootPath);
 		this.versions = new Map<string, number>();
@@ -201,7 +201,7 @@ export class ProjectManager implements Disposable {
 		const uris = await this.remoteFs.getWorkspaceFiles(util.path2uri('', rootPath));
 		for (const uri of uris) {
 			const file = util.uri2path(uri);
-			const rel = path_.posix.relative(this.rootPath, util.normalizePath(file));
+			const rel = path_.posix.relative(this.rootPath, util.toUnixPath(file));
 			if (!moduleStructureOnly || util.isGlobalTSFile(rel) || util.isConfigFile(rel) || util.isPackageJsonFile(rel)) {
 				filesToFetch.push(file);
 			} else if (!this.localFs.fileExists(rel)) {
@@ -269,7 +269,7 @@ export class ProjectManager implements Disposable {
 			for (const uri of uris) {
 				const file = util.uri2path(uri);
 				if (
-					util.normalizePath(file).indexOf('/node_modules/') === -1
+					util.toUnixPath(file).indexOf('/node_modules/') === -1
 					&& (util.isJSTSFile(file) || util.isConfigFile(file) || util.isPackageJsonFile(file))
 				) {
 					filesToEnsure.push(file);
@@ -318,7 +318,7 @@ export class ProjectManager implements Disposable {
 	 */
 	ensureFilesForReferences(uri: string): Promise<void> {
 		const fileName: string = util.uri2path(uri);
-		if (util.normalizePath(fileName).indexOf(`${path_.posix.sep}node_modules${path_.posix.sep}`) !== -1) {
+		if (util.toUnixPath(fileName).indexOf(`${path_.posix.sep}node_modules${path_.posix.sep}`) !== -1) {
 			return this.ensureFilesForWorkspaceSymbol();
 		}
 
@@ -345,7 +345,7 @@ export class ProjectManager implements Disposable {
 		}
 		filePaths.forEach(f => seen.add(f));
 
-		const absFilePaths = filePaths.map(f => util.normalizePath(util.resolve(this.rootPath, f)));
+		const absFilePaths = filePaths.map(f => util.toUnixPath(util.resolve(this.rootPath, f)));
 		await this.ensureFiles(absFilePaths);
 
 		if (maxDepth > 0) {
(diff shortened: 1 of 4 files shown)


src/typescript-service.ts
```diff
@@ -69,7 +69,7 @@ export class TypeScriptService implements LanguageHandler {
 
 	async initialize(params: InitializeParams, token = CancellationToken.None): Promise<InitializeResult> {
 		if (params.rootUri || params.rootPath) {
-			this.root = params.rootPath || util.uri2path(params.rootUri);
+			this.root = params.rootPath ? util.uri2path(params.rootPath) : util.uri2path(params.rootUri);
```
felixfbecker: `rootPath` is a path, not a URI

src/util.ts
```diff
@@ -38,7 +38,7 @@ export function docstring(parts: ts.SymbolDisplayPart[]): string {
  * Normalizes path to match POSIX standard (slashes)
  */
 export function normalizePath(file: string): string {
-	return file.replace(new RegExp('\\' + path.sep, 'g'), path.posix.sep);
+	return file.replace(new RegExp('[\/\\\\]', 'g'), path.posix.sep);
```
felixfbecker: Unix allows backslashes as valid characters in file names, so this is incorrect
```
$ touch \\.txt
$ ls -la
-rw-r--r--     1 felix  staff       0 Mar 15 10:32 \.txt
```

Also equivalent to `return file.replace(/\\/g, '/')` (no need to replace forward slashes with forward slashes, no need to use the regexp constructor and `path.posix.sep` is always the forward slash)

src/test/typescript-service-helpers.ts
```diff
@@ -1577,12 +1577,15 @@ export function describeTypeScriptService(TypeScriptServiceConstructor = TypeScr
...
+			['file:///special-characters-in-path/%40foo/b.ts', 'export function b() {}'],
+			['file:///windows/app/master.ts', '/// <reference path="..\\lib\\master.ts" />\nc();'],
+			['file:///windows/lib/master.ts', '/// <reference path="..\\lib\\slave.ts" />'],
+			['file:///windows/lib/slave.ts', 'function c() {}']
 		])));
 
 		afterEach(<any> shutdownTypeScriptService);
 
-		describe('getHovers()', <any> function (this: TestContext) {
+		describe('LSP server', <any> function (this: TestContext) {
```
felixfbecker: why this rename? I like to have this structure
- describe TypeScriptService
  - describe specific workspace context
    - describe method
       - it should ...

src/project-manager.ts
```diff
@@ -357,7 +357,7 @@ export class ProjectManager implements Disposable {
 				const info = ts.preProcessFile(contents, true, true);
 				const compilerOpt = config.getHost().getCompilationSettings();
 				for (const imp of info.importedFiles) {
-					const resolved = ts.resolveModuleName(imp.fileName, filePath, compilerOpt, config.moduleResolutionHost());
+					const resolved = ts.resolveModuleName(util.normalizePath(imp.fileName), filePath, compilerOpt, config.moduleResolutionHost());
```
felixfbecker: So `preProcessFile()` will return raw, relative, maybe-Windows imports, but `resolveModuleName()` only takes relative Posix paths?
Given the problem I commented further down, how about we first try to resolve the raw path, then if it didn't resolve, try again with backslashes replaced with forward slashes?

src/typescript-service.ts
alexsaveliev: Duplicates https://github.com/sourcegraph/sourcegraph/issues/4542

src/util.ts
alexsaveliev: Not sure if TypeScript does distinguish them in triple slash references

src/test/typescript-service-helpers.ts
alexsaveliev: It makes 
* TypeScriptService
  * Special file names
    * LSP server
        * should accept files with ...

felixfbecker: Yep, but `LSP server` is redundant / incorrect. The unit under test here is the `TypeScriptService` class, specifically the `getHover()` method, in the context of a workspace with files with special names. The description was correct before

src/project-manager.ts
alexsaveliev: `resolveModuleName` will use our implemenation of `ModuleResolutionHost`

src/test/typescript-service-helpers.ts
alexsaveliev: Well, with the addition of new test scenario `getHover()` is no longer reflects properly what we are testing. New test case uses `getDefinition` because `getHover()` does not provide enough proof of correctness in the triple-slash references case.
Also, frankly I don't care what we are using to test server behavior when dealing with not-common file paths, my goal is just to test that we can handle them

felixfbecker: Okay, then please remove the `LSP server` describe block

src/typescript-service.ts
felixfbecker: But that is an error in the sourcegraph implementation, which is why I moved the workaround to the build server. I would like to keep it out of here because this is the public LS that is used by other clients that behave correctly

alexsaveliev: this addition does not break communication with other clients who sent proper data, but allows me to debug TSJS LS server in combination with Sourcegraph w/o need to run other components.
uri2path does not change string unless it looks like uri, so
`uri2path('file:///foo') = '/foo'`
`uri2path('/foo') = '/foo'`

alexsaveliev: removed this change

src/util.ts
alexsaveliev: removed this change

src/project-manager.ts
```diff
@@ -357,7 +357,7 @@ export class ProjectManager implements Disposable {
 				const info = ts.preProcessFile(contents, true, true);
 				const compilerOpt = config.getHost().getCompilationSettings();
 				for (const imp of info.importedFiles) {
-					const resolved = ts.resolveModuleName(imp.fileName, filePath, compilerOpt, config.moduleResolutionHost());
+					const resolved = ts.resolveModuleName(imp.fileName.replace(/\\/g, '/'), filePath, compilerOpt, config.moduleResolutionHost());
```
felixfbecker: I think `normalizePath` should always replace all backslashes like this here, independent of the host OS - I cannot think of a use case where the host OS should dictate anything, can you?
Maybe the function should just be named `toPosixPath` or something like that.

PR #103, opened 2017-03-18
Add advanced logging through LSP

### LINKED ISSUE #72: Log through LSP, not STDERR
I would like to handle logging through a passed Logger interface that logs through LSP instead of STDERR.

felixfbecker:
Closes #72 

- Adds a `Logger` interface and implementations over STDIO and LSP
- Removes all `console.error()` occurrences and replaces with the appropriate log calls
- Logs by the TCP cluster are done to STDIO, everything else over LSP
- Workers use a `PrefixedLogger`, no more global console rewrite
- `--logfile` uses a `FileLogger`


src/connection.ts
```diff
@@ -1,44 +1,34 @@
...
-		listen(tracer);
+		});
 	};
 
 	const writer = new StreamMessageWriter(output);
 	const write = writer.write.bind(writer);
 	writer.write = (message: Message) => {
-		doTrace(message, trace, logger, '<--');
+		if (options.trace) {
+			logger.log('-->', message);
```
alexsaveliev: Please reverse arrow direction

felixfbecker: woops, thanks!

PR #108, opened 2017-03-20
Fetch once

### LINKED ISSUE #96: File content is fetched multiple times
When adding a list of fetched content to `RemoteFileSystem.getTextDocumentContent()` and logging when content is fetched twice, you can see that apparently every single file is fetched at least twice (without having done any request besides `initialize`).

My suspicion is that this is because on `initialize` `ensureFilesForWorkspaceSymbols()` is called, whose result promise is memoized, but other places in the code (`ensureFilesForHoverAndDefinition()`? `ProjectConfiguration.init()`?) fetch them directly again.

### Solution
Eventually, these file contents will be added to the `InMemoryFileSystem`. This should be the source of truth for file contents and whether they are loaded or not. One difficulty I see here is the "dummy" placeholder that is added for some files (ideally this should be removed).

I want `InMemoryFileSystem` to be able to hold the workspace structure (as returned from `workspace/files`), but only _optionally_ hold a content for each file.
Additional benefit: `RemoteFileSystem.getWorkspaceFiles()` is currently memoized, instead should only be used to hydrate the structure of `InMemoryFileSystem` once (that is, *.ts, tsconfig.json, package.json), contents in the background and then all parts of the codebase should just interact with that (and invalidate / modify it as needed).

### TODO

- [x] Extend `InMemoryFileSystem` to differentiate whether a file does not _exist_ or is just not _loaded_ 
- [ ] Find out which functions call the remote file system to fetch files directly
- [ ] Adapt them
- [x] Remove memoization in `RemoteFileSystem` and hydrate the structure once on initialization
- [ ] Make `InMemoryFileSystem` available to `TypeScriptService` and the build handler, remove direct usage of `RemoteFileSystem` (for other purposes than hydration)

### Difficulties
- ~~`ProjectManager`'s, `ProjectConfiguration`'s and `InMemoryFileSystem`'s properties, methods and parameters are undocumented and I have a hard time understanding the semantics of each one~~
- ~~`InMemoryFileSystem.entries` is just `any`. It's hard for me to understand how the tree is typed.~~

wdyt @beyang @alexsaveliev

felixfbecker:
Closes #96 

This adds a new class, `FileSystemUpdater`. It handles syncing the remote file system with the in-memory file system and provides four methods: one for syncing structure, one for syncing content, and memoized versions of them.
Memoization is done by a new memoization helper that supports cancellation (it's complicated, but tested).

All syncing goes through this class, eventually I want to remove the `ensure*` methods from the ProjectManager and make every method use this class, but that is out of scope for this PR, for now I just adapted them.

The InMemoryFileSystem is extended to also have a dumb list of files that are _available_ in the workspace, but not necessarily fetched. Many parts of the server just iterate over the files and don't have to know the content of each file.

I recommend to review commit by commit.

Need to fix Windows tests, probably a `path.` that should have been `path.posix.`


src/fs.ts
```diff
@@ -82,3 +74,95 @@ export class LocalFileSystem implements FileSystem {
...
+	get(uri: string | undefined): Promise<void> | undefined {
+		let hit = super.get(uri);
+		if (hit) {
+			return hit;
+		}
+		// Find out if parent folder is being fetched already
+		hit = super.get(undefined);
+		if (hit) {
+			return hit;
+		}
```
alexsaveliev: Why `undefined` has precedence over parent URI?

src/project-manager.ts
```diff
@@ -733,11 +721,36 @@ export class InMemoryFileSystem implements ts.ParseConfigHost, ts.ModuleResoluti
...
-	 * @param path relative file path
-	 * @param content file content
+	 *
+	 * @param path File path, absolute or relative to rootPath
+	 * @param content File content
 	 */
 	addFile(path: string, content: string) {
+		// Ensure path is relative to rootpath
+		if (path_.posix.isAbsolute(path)) {
+			path = path_.relative(this.path, path);
```
alexsaveliev: Mix of `path._posix` and `path` looks a little bit suspicious

src/fs.ts
felixfbecker: `undefined` means no `base` was provided, in which case it defaults to the workspace root according to LSP

PR #109, opened 2017-03-21
Strict null checks

alexsaveliev:
- enabling `strictNullCheck`


src/project-manager.ts
```diff
@@ -761,14 +761,18 @@ export class InMemoryFileSystem implements ts.ParseConfigHost, ts.ModuleResoluti
...
 	}
 
 	/**
 	 * @param path file path (both absolute or relative file paths are accepted)
-	 * @return file's content in the following order (overlay then cache) if any
+	 * @return file's content in the following order (overlay then cache)
 	 */
-	readFile(path: string): string | undefined {
+	readFile(path: string): string {
+		return this.readFileImpl(path) || '';
```
felixfbecker: This looks like something that could cause hard to track bugs. Would it work to throw an error instead?

src/project-manager.ts
```diff
@@ -761,14 +761,18 @@ export class InMemoryFileSystem implements ts.ParseConfigHost, ts.ModuleResoluti
...
+	}
+
+	private readFileImpl(path: string): string | undefined {
```
felixfbecker: Name this something like `readFileIfExists` and document the difference to `readFile`

src/fs.ts
```diff
@@ -71,7 +71,7 @@ export class LocalFileSystem implements FileSystem {
 		return path.resolve(this.rootPath, uri2path(uri));
 	}
 
-	async getWorkspaceFiles(base?: string): Promise<string[]> {
+	async getWorkspaceFiles(base: string): Promise<string[]> {
```
felixfbecker: This must stay optional, I fixed the bug in my other branch:
https://github.com/sourcegraph/javascript-typescript-langserver/pull/108/files#diff-a987912cd3148bf666f759bdbf896718R66

src/typescript-service.ts
```diff
@@ -69,14 +69,14 @@ export class TypeScriptService implements LanguageHandler {
 
 	async initialize(params: InitializeParams, token = CancellationToken.None): Promise<InitializeResult> {
 		if (params.rootUri || params.rootPath) {
-			this.root = params.rootPath || util.uri2path(params.rootUri);
+			this.root = params.rootPath || util.uri2path(params.rootUri as string);
```
felixfbecker: To assert that a property exists, suffix with an exclamation mark

src/util.ts
```diff
@@ -213,7 +213,7 @@ export function symbolDescriptorMatch(query: rt.PartialSymbolDescriptor, sym: rt
 			continue;
 		}
 		if (key === 'package') {
-			if (!sym.package || !packageDescriptorMatch(query.package, sym.package)) {
+			if (!sym.package || !packageDescriptorMatch(query.package as rt.PackageDescriptor, sym.package)) {
```
felixfbecker: why this cast? Does `packageDescriptorMatch` accept `undefined` as possible value?

felixfbecker:
> fixed appveyor integration (I found no wget in Git directory but found it in C:\msys64, using it)

what do you mean with that? Git directory is added to get `bash` in PATH - what was broken?

alexsaveliev:
> what do you mean with that? Git directory is added to get bash in PATH - what was broken?

Seems so, at least for me:
https://ci.appveyor.com/project/alexsaveliev/javascript-typescript-langserver/build/44

felixfbecker:
Maybe they changed their bash script to now use `curl`. If it works, 👍

src/project-manager.ts
alexsaveliev: Won't work, sorry. There might be a case when TS compiler requests file that does not exists and if we'll throw an error it may break source analysis

src/project-manager.ts

src/fs.ts
alexsaveliev: done

src/typescript-service.ts
alexsaveliev: done

src/util.ts
alexsaveliev: replaced with an exclamation mark

Issue #115, opened 2017-03-27
Add Tracing


felixfbecker:
Implement https://github.com/opentracing/opentracing-javascript

- Create a span at every LSP method call, pass it to functions
- Create, finish child spans for stuff like fetching files, compilation
- Make it compatible with https://www.npmjs.com/package/lightstep-tracer

Issue #106, opened 2017-03-19
Make code more resolved around URIs, not file paths


felixfbecker:
LSP uses URIs to identify files. The LS uses file paths almost exclusively. This makes it impossible to, for example, to identify files that are not saved yet in the editor. It is not guaranteed that a URI can always be translated to a file path.

The TS language service probably only works with paths, but that translation should happen as late as possible, not as early as possible. In practice this means methods should take URIs where possible and classes like `InMemoryFileSystem` must be able to save files by URI.

PR #120, opened 2017-03-27
Expose updater

felixfbecker:
A couple of changes:
- Optimize some iterations by using Iterators
- Remove cancellation from file fetching for now. It is too complex to get right with regard to memoization with little benefit and would complicate #115 
- Expose the `FileSystemUpdater` as a protected property of the `TypeScriptService` and remove `beforeProjectInit` hook in favor of `initializeFileSystems`
- Expose some memoization caches so they can be cleared
- Correct logging
- Save all file contents in the `files` URI -> content map. All places where the paths are used were adapted under the assumption that paths and URIs can be converted back and forth to `file://` URLs. This is a step towards #106

### DIFF
--- a/src/fs.ts
+++ b/src/fs.ts
@@ -3,6 +3,7 @@ import * as path from 'path';
 import { cancellableMemoize, CancellationToken, throwIfRequested } from './cancellation';
 import { LanguageClientHandler } from './lang-handler';
 import glob = require('glob');
+import iterate from 'iterare';
 import Semaphore from 'semaphore-async-await';
 import * as url from 'url';
 import { InMemoryFileSystem } from './project-manager';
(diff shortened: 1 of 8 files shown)


src/fs.ts
```diff
@@ -146,23 +126,42 @@ export class FileSystemUpdater {
...
-		}
+	fetchStructure(): Promise<void> {
+		const promise = (async () => {
+			try {
+				const uris = await this.remoteFs.getWorkspaceFiles();
+				for (const uri of uris) {
+					this.inMemoryFs.add(uri);
+				}
+			} catch (err) {
+				this.structureFetch = undefined;
```
alexsaveliev: I'd suggest to call `invalidateStructure` here

PR #121, opened 2017-03-28
Yet another step to cross-repo j2d

alexsaveliev:
- fixed package extraction: package.json may be located at the upper level of tsconfig.json
- corrected container name extraction: we should skip name component (if any), so we are taking last part after dot and if it contains path separators then we consider that container name is empty

### DIFF
--- a/src/project-manager.ts
+++ b/src/project-manager.ts
@@ -1015,11 +1015,19 @@ export class ProjectConfiguration {
 	 * @return package name (project name) of a given project
 	 */
 	getPackageName(): string | null {
-		const pkgJsonFile = path_.posix.join(this.rootFilePath, 'package.json');
-		if (this.fs.fileExists(pkgJsonFile)) {
-			return JSON.parse(this.fs.readFile(pkgJsonFile)).name;
+		// package.json may be located at the upper level as well
+		let currentDir = this.rootFilePath;
+		while (true) {
+			const pkgJsonFile = path_.posix.join(currentDir, 'package.json');
+			if (this.fs.fileExists(pkgJsonFile)) {
+				return JSON.parse(this.fs.readFile(pkgJsonFile)).name;
+			}
+			const parentDir = path_.dirname(currentDir);
+			if (parentDir === '.' || parentDir === '/' || parentDir === currentDir) {
+				return null;
+			}
+			currentDir = parentDir;
 		}
-		return null;
 	}
 
 	/**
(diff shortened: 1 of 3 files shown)


src/util.ts
```diff
@@ -249,3 +249,13 @@ function lastDotCmp(s: string): string {
...
 	return cmps[cmps.length - 1];
 }
+
+/**
+ * Strips file part (if any) from container name (last component of container path)
+ * For example TS may return the following name: /node_modules/vscode-jsonrpc/lib/cancellation.
+ * We consider that if name contains path separtor then container name is empty
+ * @param containerName
+ */
+function stripFileInfo(containerName: string): string {
```
felixfbecker: Could you explain why this is needed?
Iirc the `containerName` is used inside `SymbolDescriptor`, so changing it affects the uniqueness of it

src/project-manager.ts
```diff
@@ -1015,11 +1015,19 @@ export class ProjectConfiguration {
...
-		if (this.fs.fileExists(pkgJsonFile)) {
-			return JSON.parse(this.fs.readFile(pkgJsonFile)).name;
+		// package.json may be located at the upper level as well
+		let currentDir = this.rootFilePath;
+		while (true) {
+			const pkgJsonFile = path_.posix.join(currentDir, 'package.json');
+			if (this.fs.fileExists(pkgJsonFile)) {
+				return JSON.parse(this.fs.readFile(pkgJsonFile)).name;
+			}
+			const parentDir = path_.dirname(currentDir);
```
felixfbecker: not posix?

alexsaveliev: For `dirname` OS-specific `path` is enough

src/util.ts
alexsaveliev: When we looking at symbol in project it's container name may look like `/node_modules/path-to/file` but when we are searching this symbol in it's own repo container name is empty (top-level symbol). This making two symbols not match when we are trying to resolve cross-repo symbol location. I think we need to process container name as returned by TS the following way:
- extract last component (after last dot if any)
- check if there are path separators, if there are - I assume we are looking at top-level element defined in some file so container name is empty string

felixfbecker: I think you are right, we cannot guarantee that a file defining the symbol corresponds to the file referencing the symbol in any way: For TS projects, it can be in a different folder (src vs lib) and it is a different file (.ts vs .d.ts vs .js). For JS projects, the typings are probably in a single index.d.ts on DefinitelyTyped.

Leaving the path out will make symbol matching very fuzzy, but I don't see a different way.

Issue #129, opened 2017-04-04
permission denied, scandir '/lost+found'


mickaelistria:
I'm trying to connect Eclipse LSP4E with this Language Server.
I've run the build instructions without issues, and LSP4E It basically starts `node /home/mistria/git/javascript-typescript-langserver/lib/language-server-stdio.js` and connects to stdio streams.

The server says it's failing because of some `permission denied, scandir '/lost+found'`, see full conversation below.

```
!ENTRY org.eclipse.lsp4e 1 0 2017-04-04 19:01:19.852
!MESSAGE RequestMessage
{
  "jsonrpc": "2.0",
  "id": "1",
  "method": "initialize",
  "params": {
    "rootUri": "file:/home/mistria/git/mathjs/",
    "capabilities": {
      "workspace": {
        "applyEdit": true,
        "symbol": {},
        "executeCommand": {}
      },
      "textDocument": {
        "synchronization": {
          "willSave": true,
          "willSaveWaitUntil": true,
          "didSave": true
        },
        "completion": {
          "completionItem": {
            "snippetSupport": true
          }
        },
        "hover": {},
        "signatureHelp": {},
        "references": {},
        "documentHighlight": {},
        "documentSymbol": {},
        "formatting": {},
        "rangeFormatting": {},
        "definition": {},
        "codeAction": {},
        "codeLens": {},
        "documentLink": {},
        "rename": {}
      }
    },
    "clientName": "Eclipse Platform"
  }
}
glob error { Error: EACCES: permission denied, scandir '/lost+found'
  errno: -13,
  code: 'EACCES',
  syscall: 'scandir',
  path: '/lost+found' }

!ENTRY org.eclipse.lsp4e 1 0 2017-04-04 19:01:20.598
!MESSAGE ResponseMessage
{
  "jsonrpc": "2.0",
  "id": "1",
  "result": {
    "capabilities": {
      "textDocumentSync": 1,
      "hoverProvider": true,
      "completionProvider": {
        "resolveProvider": false,
        "triggerCharacters": [
          "."
        ]
      },
      "definitionProvider": true,
      "referencesProvider": true,
      "documentSymbolProvider": true,
      "workspaceSymbolProvider": true
    }
  }
}

!ENTRY org.eclipse.lsp4e 1 0 2017-04-04 19:01:20.608
!MESSAGE NotificationMessage
{
  "jsonrpc": "2.0",
  "method": "textDocument/didOpen",
  "params": {
[...]

Issue #114, opened 2017-03-25
hover/refresh results in access error


aleksandersumowski:
I'm on a Mac, I'm using neovim with /autozimu/LanguageClient-neovim/. When attempting to eg textDocument/definition I get the following error message in response. Any suggestions? Directory /Library/Application Support/ApplePushService is only accessible as a root.
```json
{"code": -32603, "message": "Request textDocument/definition failed with message: EACCES: permission denied, scandir '/Library/Application Support/ApplePushService'"}
```

felixfbecker:
What version of the LS are you running and what is the rootPath / rootUri being sent?

aleksandersumowski:
You mean version of javascript-typescript-langserver? It's 1.2

How can I check rootPath / rootUri being sent?

alexsaveliev:
@aleksandersumowski 

> How can I check rootPath / rootUri being sent?

Please run the server with `-t` to log requests/responses to console or with `-l file` to log them to a file. There should be `initialize` request which has either `rootPath` or `rootUri` parameter

Thanks!

aleksandersumowski:
For the neovim plugin turns out the simple way to get debug information was to:
`:call LanguageClient_setLoggingLevel('DEBUG')`
before calling
`:LanguageClientStart`

The resulting log is:

23:10:08 INFO     Begin LanguageClientStart
23:10:08 INFO     End LanguageClientStart
23:10:08 INFO     Begin initialize
23:10:08 INFO     rootPath: /Users/aleksander.sumowski/projects/broadband-web
23:10:08 DEBUG     => {"jsonrpc": "2.0", "method": "initialize", "params": {"processId": 32092, "rootPath": "/Users/aleksander.sumowski/projects/broadband-web", "rootUri": "file:///Users/aleksander.sumowski/projects/broadband-web", "capabilities": {}}, "id": 0}
23:10:09 DEBUG     <= {"jsonrpc":"2.0","id":0,"result":{"capabilities":{"textDocumentSync":1,"hoverProvider":true,"definitionProvider":true,"referencesProvider":true,"documentSymbolProvider":true,"workspaceSymbolProvider":true,"xworkspaceReferencesProvider":true,"xdefinitionProvider":true,"xdependenciesProvider":true,"completionProvider":{"resolveProvider":false,"triggerCharacters":["."]},"xpackagesProvider":true}}}
23:10:09 INFO     End initialize
[...]

PR #131, opened 2017-04-04
using vscode-uri to transform URI to path and in back

alexsaveliev:
- using vscode-uri to transform URI to path and in back
- `textDocument/did...` update content by relative path
- I expect that this PR should deal with #114 and #129

### DIFF
--- a/src/typescript-service.ts
+++ b/src/typescript-service.ts
@@ -1362,7 +1362,7 @@ export class TypeScriptService implements LanguageHandler {
 	didOpen(params: DidOpenTextDocumentParams): Promise<void> {
 		const uri = util.uri2reluri(params.textDocument.uri, this.root);
 		return this.projectManager.ensureFilesForHoverAndDefinition(uri).then(() => {
-			this.projectManager.didOpen(util.uri2path(uri), params.textDocument.text);
+			this.projectManager.didOpen(util.uri2relpath(uri, this.root), params.textDocument.text);
 		});
 	}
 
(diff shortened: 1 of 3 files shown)


src/typescript-service.ts
```diff
@@ -1362,7 +1362,7 @@ export class TypeScriptService implements LanguageHandler {
 	didOpen(params: DidOpenTextDocumentParams): Promise<void> {
 		const uri = util.uri2reluri(params.textDocument.uri, this.root);
 		return this.projectManager.ensureFilesForHoverAndDefinition(uri).then(() => {
-			this.projectManager.didOpen(util.uri2path(uri), params.textDocument.text);
+			this.projectManager.didOpen(util.uri2relpath(uri, this.root), params.textDocument.text);
```
felixfbecker: I would prefer an explicit call to `path.relative()` if it's equivalent

src/util.ts
```diff
@@ -69,29 +71,21 @@ export function convertStringtoSymbolKind(kind: string): SymbolKind {
...
-	}
 	let p;
 	if (root) {
 		p = resolve(root, file);
 	} else {
 		p = file;
 	}
-	p = toUnixPath(p).split('/').map(encodeURIComponent).join('/');
-	return ret + p;
+	return Uri.file(p).toString();
```
felixfbecker: VS Code's URI module is broken in a couple ways. For example, it encodes `file://c:/whatever` as `file://c%3a/whatever`, while node's `url.format()` behaves correctly. I would prefer not to use it as a dependency. What is the difference in output here?

alexsaveliev:
Closing this one, needs to be redone

PR #133, opened 2017-04-05
Updated symbol kind mapping

alexsaveliev:
- returning 'variable' by default
- difference from VS code (enum members and constants are displayed as constants, not as variables)


felixfbecker:
What is the difference to VS Code you mentioned?

alexsaveliev:
difference (known) is for constants and enum members. VS code uses `variable`, we are using `constant` symbol kind. 

VS Code:
vscode-sym

Sourcegraph:
sg-sym

src/util.ts
```diff
@@ -43,28 +43,39 @@ export function toUnixPath(filePath: string): string {
...
+		// case 'keyword'
+		// case 'type'
+		// case 'call'
+		// case 'index'
+		// case 'construct'
+		// case 'primitive type'
+		// case 'label'
+		// case 'directory'
+		// case 'external module name'
+		// case 'external module name'
```
alexsaveliev: because there are textual constants in typescript code but no matching symbol kind (at least I can't think of matching one) in LSP

felixfbecker: That's okay, but they should be mapped to _something_, so lets not comment out the cases

alexsaveliev: they are mapped by the `default` statement to `variable`

Issue #119, opened 2017-03-27
Consolidate ensureFilesFor*


felixfbecker:
Having a method that has the consumer in the name is a code smell because a function should not be coupled to its consumer. Current inconsistencies:
 - we have `ensureFilesForHoverAndDefinition`, which is also called for `getCompletions`
-  there is `ensureFilesForWorkspaceSymbols`, which fetches _more_ files than `ensureAllFiles`.
- `refreshFileTree` is used for refreshing everything or only module structure depending on parameter, but does not share the same memoization cache with eg `ensureAllFiles`, so potentially causes inconsistent state
- Memoization caches need to be cleared for every method individually and explicitly, which creates a lot of coupling.

Lets identify the real use cases here:
- Fetching the whole workspace for references, symbols and general pre-fetching
  Solution: ensure file structure with `updater`, iterate over URIs, ensure each
- Fetching only a file and its references for hover, definition, completions and potentially more
  Solution: `ensureTransitiveFileDependencies`
- Fetching only module structure as part of the previous point
  Solution: `ensureModuleStructure`

In addition: The memoization caches need to be cleared when the workspace changed (eg dependencies installed). Explicitly calling a method for this is not good because it introduces a lot of coupling. An evented model would be better, where `InMemoryFileSystem` emits events when something changed and eg the ProjectManager can add a configuration if a tsconfig.json was added.

PR #134, opened 2017-04-05
Add Tracing

felixfbecker:
Pretty big PR, let's dive in.

Tracing is done with OpenTracing, I depend on my fork that rewrites in TS (PR open and approved in general). Tracing context can be sent through a custom `meta` field in the JSON RPC message. To extract it, I need to get access to the raw message - something the vscode-languageserver module hides. So I decoupled our code completely from vscode-languageserver, only use vscode-jsonrpc (which ironically also handles the LSP transport) and handle the dispatching in `registerLanguageHandler`. After all, vscode-languageserver was difficult to work with anyways, because we had to wrap it to register our class based handler, it tried to kill the process on exit, etc etc.

The new dispatcher automatically calls a method on the given target object by camelCasing the method. That means `textDocument/definition` becomes `textDocumentDefinition`. Private methods in the handler are underscore-prefixed so they are not exposed over RPC (`camelCase()` will remove any underscores).
The same naming scheme is applied to the RemoteLanguageClient too. I wrapped the vscode-jsonrpc MessageReader in a NodeJS EventEmitter to support multiple listeners.
The new dispatcher also allows us to customize a lot more too, logging now doesn't require monkey-patching anymore. Instead of vscode's CancelTokens, which showed to be very hard to compose, I want to slowly move to Observables instead of promises, which can be easily canceled and composed (it will also lay the path for implementation of `$/partialResult`).

The actual tracing is done by passing an optional parent span as the last argument to most functions. By default, a noop OpenTracing span is used. The dispatcher will create and finish a span for the RPC handlers, long-running functions can then create a child span and finish it in a `finally`. `ensure*` functions need create a span decoupled from the underlying `fetch*` call.

The Tracer to use is passed in the top-level file that hosts the CLI interface and instantiates the classes, and can be any OpenTracing-compatible tracer.

Right now only a basic set of functions are traced. I would like to add more granularity:
- trace `ensureFilesFor*` methods, might require a bit of #119. Primarily, functions should only take a single file, not multiple, otherwise it becomes harder to instrument the method.
- trace compilation, not sure where to add exactly

ref #115


src/connection.ts
```diff
@@ -1,71 +1,184 @@
...
+			observable = Observable.of(returnValue);
+		}
+		if (isRequestMessage(message)) {
+			// If request, subscribe to result and send a response
+			const subscription = observable
+				.finally(() => {
+					// Finish span
+					span.finish();
+					// Delete subscription from Map
+					subscriptions.delete(message.id);
```
keegancsmith: is it possible that this could run before `subscriptions.set(message.id, subscription);` below?

felixfbecker:
@keegancsmith There are no typings for OpenTracing, and before I write typings and do a PR to DT, I might as well just rewrite OpenTracing, because I think it's a great improvement in general for the OpenTracing reference implementation to have strictly declared types. I proposed this on Gitter and did a PR which received positive feedback, so I'm pretty confident it will be merged soon. 
 https://github.com/opentracing/opentracing-javascript/pull/68

Chaing `Fields` to `Options` was actually a requested change: https://github.com/opentracing/opentracing-javascript/pull/68#discussion_r109313347

src/connection.ts
felixfbecker: Good question.

src/language-server-stdio.ts
```diff
@@ -17,11 +19,25 @@ program
...
-registerLanguageHandler(connection, new TypeScriptService(new RemoteLanguageClient(connection), {
-	strict: program.strict
-}));
-connection.listen();
+
+registerLanguageHandler(
+	messageEmitter,
+	messageWriter,
+	service,
+	program.trace ? logger : undefined
```
alexsaveliev: I'm not sure that this operator is correct. I'd rather pass not null logger object but control if messages should be logged.

src/project-manager.ts
```diff
@@ -319,49 +331,60 @@ export class ProjectManager implements Disposable {
...
-				for (const imp of info.importedFiles) {
-					const resolved = ts.resolveModuleName(util.toUnixPath(imp.fileName), filePath, compilerOpt, config.moduleResolutionHost());
-					if (!resolved || !resolved.resolvedModule) {
-						// This means we didn't find a file defining
-						// the module. It could still exist as an
-						// ambient module, which is why we fetch
-						// global*.d.ts files.
-						continue;
+	private async ensureTransitiveFileDependencies(filePaths: string[], maxDepth: number, seen = new Set<string>(), childOf = new Span()): Promise<void> {
+		const span = childOf.tracer().startSpan('Ensure file imports', { childOf });
```
alexsaveliev: not sure if it's good idea to trace these calls, there may be plenty. Won't we pollute logs?

src/typescript-service.ts
```diff
@@ -135,26 +143,27 @@ export class TypeScriptService implements LanguageHandler {
 	 *
 	 * @param accessDisk Whether the language server is allowed to access the local file system
 	 */
-	protected initializeFileSystems(accessDisk: boolean): void {
+	protected _initializeFileSystems(accessDisk: boolean): void {
```
alexsaveliev: why `_` prefix?
Frankly I don't like this statement: 
> Private methods in the handler are underscore-prefixed so they are not exposed over RPC (camelCase() will remove any underscores).

src/connection.ts
```diff
@@ -1,71 +1,197 @@
...
-	connection.onWorkspaceSymbol((params, token) => handler.getWorkspaceSymbols(params, token));
-	connection.onRequest('workspace/xreferences', (params, token) => handler.getWorkspaceReference(params, token));
-	connection.onRequest('workspace/xpackages', (params, token) => handler.getPackages(params, token));
-	connection.onRequest('workspace/xdependencies', (params, token) => handler.getDependencies(params, token));
+export function registerLanguageHandler(
+	messageEmitter: MessageEmitter,
+	messageWriter: MessageWriter,
+	handler: TypeScriptService,
+	logger: Logger = new NoopLogger(),
+	tracer = new Tracer()
```
alexsaveliev: If I understand this code does nothing for now, right? We don't have plugged tracer?

tsconfig.json
```diff
@@ -2,7 +2,8 @@
   "compilerOptions": {
     "target": "es2016",
     "lib": [
-      "es2016"
+      "es2016",
+      "dom"
```
alexsaveliev: Why do we need `dom`?

src/connection.ts
felixfbecker: I initially had the logger only for logging the messages so you could just pass Noop for that. Will add option

src/connection.ts
```diff
@@ -1,71 +1,197 @@
...
+): void {
+	/** Tracks Subscriptions for results to unsubscribe them on $/cancelRequest */
+	const subscriptions = new Map<string | number, Subscription>();
+	messageEmitter.on('message', async message => {
+		logger.log('-->', message);
```
felixfbecker: I initially had the logger only for logging the messages so you could just pass Noop for that. Will add option

src/connection.ts
```diff
@@ -1,71 +1,197 @@
...
+			} as ResponseMessage);
+			return;
+		}
+		// If message is request and has tracing metadata, extract the span context and create a span for the method call
+		let span = new Span();
+		if (hasMeta(message) && isRequestMessage(message)) {
+			const context = tracer.extract(FORMAT_TEXT_MAP, message.meta);
+			if (context) {
+				span = tracer.startSpan('Handle ' + message.method, { childOf: context });
+				span.setTag('params', inspect(message.params));
```
felixfbecker: Yeah, but it's intentional here. The tag values are shown in a single line, so limiting the depth makes sense.

src/connection.ts
```diff
@@ -1,71 +1,197 @@
...
+						id: message.id,
+						error: {
+							message: err.message + '',
+							code: typeof err.code === 'number' ? err.code : ErrorCodes.InternalError,
+							data: omit(err, ['message', 'code'])
+						}
+					} as ResponseMessage);
+					// Set error on span
+					span.setTag('error', true);
+					span.log({ 'event': 'error', 'error.object': err });
```
felixfbecker: The error is returned to the client, which can log the error.

src/connection.ts
alexsaveliev: Maybe JSON?

felixfbecker: `inspect` is more readable

src/fs.ts
```diff
@@ -127,24 +129,42 @@ export class FileSystemUpdater {
...
 	 */
-	ensure(uri: string): Promise<void> {
-		return this.fetches.get(uri) || this.fetch(uri);
+	async ensure(uri: string, childOf = new Span()): Promise<void> {
+		const span = childOf.tracer().startSpan('Ensure file content', { childOf });
+		span.addTags({ uri, fresh: !this.fetches.has(uri) });
+		try {
+			return await (this.fetches.get(uri) || this.fetch(uri));
+		} catch (err) {
+			span.setTag('error', true);
```
felixfbecker: yeah, thought about that too, but couldn't settle on a good API. Would leave it as this for now.

src/lang-handler.ts
```diff
@@ -1,79 +1,111 @@
...
+	 * @param method The method to call
+	 * @param params The params to pass to the method
+	 * @return Emits the value of the result field or the error
+	 */
+	private request(method: string, params: any[] | { [attr: string]: any }): Observable<any> {
+		return new Observable<any>(subscriber => {
+			// Generate a request ID
+			const id = this.idCounter++;
+			const message: RequestMessage = { jsonrpc: '2.0', method, id, params };
+			this.logger.log('<--', message);
```
felixfbecker: you can just omit the logger

src/language-server-stdio.ts
```diff
@@ -17,11 +19,25 @@ program
...
+const messageEmitter = new MessageEmitter(process.stdin);
+const messageWriter = new StreamMessageWriter(process.stdout);
+const remoteClient = new RemoteLanguageClient(messageEmitter, messageWriter);
+const service = new TypeScriptService(remoteClient, { strict: program.strict });
+
+// Add an exit notification handler to kill the process
+messageEmitter.on('message', message => {
+	if (isNotificationMessage(message) && message.method === 'exit') {
+		logger.log(`Exit notification`);
+		process.exit(0);
```
felixfbecker: This is the correct behavior for the STDIO version, an exit notification should kill the process.

src/connection.ts
alexsaveliev: Okay, got it

src/connection.ts
alexsaveliev: I'd rather have full info than readable one

src/language-server-stdio.ts
alexsaveliev: If I remember well `src` can plug langservers specified by the file path. Also, it sends chunks of initialize-request-shutdown-exit. Won't we break this behavior with the proposed change?

src/typescript-service.ts
felixfbecker: I want methods to be dispatched dynamically, which means every LSP method corresponds to a method in the handler by a well-defined naming scheme (camelCase), instead of mapping everything manually. Of course, there needs to be a way to prevent private methods from being called. Underscore-prefixing is the easiest way for this. I generally think underscore-prefixes are bad coding style too, but here they have a justification.

tsconfig.json
felixfbecker: rxjs references DOM types

src/lang-handler.ts
felixfbecker: Updated to use an options bag

src/language-server-stdio.ts
felixfbecker: Discussed on Slack

PR #135, opened 2017-04-05
Revert LocalFileSystem breaking change

alexsaveliev:
after my changes buildserver became broken because overridden `resolveUriToPath` wasn't used anymore


src/fs.ts
```diff
@@ -63,7 +63,7 @@ export class LocalFileSystem implements FileSystem {
 	}
 
 	async getWorkspaceFiles(base?: string): Promise<Iterable<string>> {
-		const root = base ? this.resolveUriToPath(base) : this.rootPath;
+		const root = this.resolveUriToPath(base || this.rootPath);
```
felixfbecker: `rootPath` is a path, but `resolveUriToPath` is documented as taking a URI
If you have a better idea for the API feel free to change it

alexsaveliev: addressed

PR #136, opened 2017-04-05
Implement signatureHelp

tomv564:
If you are taking outside contributions: This PR is a first pass at supporting textDocument/signatureHelp.
Is feature parity with Atom a requirement? 

To do:
- [x] ~~`activeSignature` and `activeParameter` should probably be undefined if `signatures` is empty~~
  - SignatureHelp type doesn't permit this, using 0 as default for now. 
  - vscode-languageserver-node will not allow undefined (see https://github.com/Microsoft/vscode-languageserver-node/issues/191)
- [x] rebase


src/test/typescript-service-helpers.ts
```diff
@@ -1445,6 +1445,93 @@ export function describeTypeScriptService(createService: TypeScriptServiceFactor
 		});
 	});
 
+	describe('getSignatureHelp()', <any> function (this: TestContext) {
+		beforeEach(<any> initializeTypeScriptService(createService, new Map([
+			['file:///a.ts', [
+				'class A {',
+				'	/** foo doc*/',
+				'    foo() {}',
+				'	/** bar doc*/',
```
felixfbecker: `@param` tags should be read

src/test/typescript-service-helpers.ts
```diff
@@ -1445,6 +1445,93 @@ export function describeTypeScriptService(createService: TypeScriptServiceFactor
...
+			assert.equal(result.activeParameter, 1);
+
+			assert.equal(activeSignature.label, 'baz(num, text): string');
+			assert.equal(activeSignature.documentation, 'baz doc');
+			assert.isArray(activeSignature.parameters);
+			if (activeSignature.parameters) {
+				assert.equal(activeSignature.parameters.length, 2);
+				assert.equal(activeSignature.parameters[0].label, 'num');
+				assert.equal(activeSignature.parameters[1].label, 'text');
+				assert.isUndefined(activeSignature.parameters[0].documentation);
```
felixfbecker: why not use `assert.deepEqual()` for assertions?

alexsaveliev:
@tomv564 thanks a lot for your contribution!

src/typescript-service.ts
```diff
@@ -113,6 +116,7 @@ export class TypeScriptService implements LanguageHandler {
 				// Tell the client that the server works in FULL text document sync mode
 				textDocumentSync: TextDocumentSyncKind.Full,
 				hoverProvider: true,
+				signatureHelpProvider: true,
```
alexsaveliev: I'd say 
```
				signatureHelpProvider: {
					triggerCharacters: ['(', ',']
				},
```
This way it works in VS code when you hitting "(" or ","

BTW found a bug in our code while played with your PR, thanks! :smile:

tomv564:
Thanks for this review and your efforts on this project!
Let me know if you would like more fixes, or if you'd like the commits squashed.

src/test/typescript-service-helpers.ts
```diff
@@ -1445,6 +1445,114 @@ export function describeTypeScriptService(createService: TypeScriptServiceFactor
...
+			assert.deepEqual(result, {
+				signatures: [
+					{
+						label: 'baz(num, text): string',
+						documentation: 'The Baz function',
+						parameters: [{
+							label: 'num',
+							documentation: 'Number parameter'
+						}, {
+							label: 'text',
```
felixfbecker: these should include the types of the parameters, like with tsserver:

image

alexsaveliev:
addressed all comments except this one:

> should definitely be undefined if LSP allows it. Passing zero is incrrect if the array is empty. If the typings are wrong, please try cast to the type or any if necessary and open an issue at vscode-languageserver :)

tomv564:
@alexsaveliev I should have pushed my rebase + fixes this morning, we did some double work :) I'll be careful not to push my commits so this branch can stay stable.

A maintainer answered on the issue in vscode-languageserver-node (https://github.com/Microsoft/vscode-languageserver-node/issues/191). Apparently the protocol was wrong and that 0 is an ok value for the empty response.

I will leave this PR in your hands, but look forward to contributing more in the future!

PR #137, opened 2017-04-06
Correct didChange handling

alexsaveliev:
- fixed issue when `didChange` is called before hover/def/... (when configurations are not initialized yet)


felixfbecker:
Could you explain the issue this fixes a bit more?

alexsaveliev:
> Could you explain the issue this fixes a bit more?

- open TS project in VS Code 
- do not invoke hover/definition/... requests, start typing text immediately (this causes `textDocument/didChange` notifications to be fired
- upon reception of such notification we are invalidating TS data, but the problem is that this data is not available - we haven't initialized it yet (which happens in hover/definition/... handlers).

PR #143, opened 2017-04-07
Correct shutdown and exit handling

### LINKED ISSUE #142: regression: invalid shutdown handler
- open vscode-sourcegraph extension project in VS code
- start debug session
- do a hover
- close 'Extension Development Host' window (it sends `shutdown`)

-> 
```
Unhandled exception has occurred: Error
Error: The received response has neither a result nor an error property.
    at handleInvalidMessage (c:\projects\vscode-sourcegraph\node_modules\vscode-jsonrpc\lib\main.js:294:40)
    at StreamMessageReader.callback (c:\projects\vscode-sourcegraph\node_modules\vscode-jsonrpc\lib\main.js:366:13)
    at StreamMessageReader.onData (c:\projects\vscode-sourcegraph\node_modules\vscode-jsonrpc\lib\messageReader.js:197:18)
```

alexsaveliev:
Closes #142 

- ~~added noop `exit` handler~~ ignore `exit`
- adjusted `shutdown` handler signature

### DIFF
--- a/src/typescript-service.ts
+++ b/src/typescript-service.ts
@@ -143,8 +143,13 @@ export class TypeScriptService {
 		this.inMemoryFileSystem = new pm.InMemoryFileSystem(this.root);
 	}
 
-	async shutdown(params = {}, span = new Span()): Promise<void> {
+	async shutdown(params = {}, span = new Span()): Promise<null> {
 		this.projectManager.dispose();
+		return null;
+	}
+
+	async exit(params = {}, span = new Span()): Promise<void> {
+		// noop
 	}
 
 	async textDocumentDefinition(params: TextDocumentPositionParams, span = new Span()): Promise<Location[]> {
--- a/src/typescript-service.ts
+++ b/src/typescript-service.ts
@@ -148,10 +148,6 @@ export class TypeScriptService {
 		return null;
 	}
 
-	async exit(params = {}, span = new Span()): Promise<void> {
-		// noop
-	}
-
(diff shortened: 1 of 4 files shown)


src/typescript-service.ts
```diff
@@ -143,8 +143,13 @@ export class TypeScriptService {
 		this.inMemoryFileSystem = new pm.InMemoryFileSystem(this.root);
 	}
 
-	async shutdown(params = {}, span = new Span()): Promise<void> {
+	async shutdown(params = {}, span = new Span()): Promise<null> {
 		this.projectManager.dispose();
+		return null;
+	}
+
+	async exit(params = {}, span = new Span()): Promise<void> {
```
felixfbecker: I would like `exit` to be handled by the part that registers the handler, because that needs to kill the process or close the socket. I would ignore `exit` in `registerLanguageHandler`

src/typescript-service.ts
```diff
@@ -143,8 +143,13 @@ export class TypeScriptService {
 		this.inMemoryFileSystem = new pm.InMemoryFileSystem(this.root);
 	}
 
-	async shutdown(params = {}, span = new Span()): Promise<void> {
+	async shutdown(params = {}, span = new Span()): Promise<null> {
 		this.projectManager.dispose();
+		return null;
```
felixfbecker: Good catch. We should only let notification handlers return `void`, and request handlers always `null` if they don't have a result.

PR #144, opened 2017-04-07
Log outgoing messages

### LINKED ISSUE #141: Responses are not logged anymore
- running `node lib/language-server.js -p 2089 -t` which supposed to trace requests and responses

-> requests are traced, responses aren't

```
DEBUG --> { jsonrpc: '2.0',
  id: 13,
  method: 'textDocument/hover',
  params:
   { textDocument: { uri: 'file:///c%3A/projects/javascript-typescript-langserve
r/src/project-manager.ts' },
     position: { line: 0, character: 9 } } }
DEBUG --> { jsonrpc: '2.0',
  id: 14,
  method: 'textDocument/hover',
  params:
   { textDocument: { uri: 'file:///c%3A/projects/javascript-typescript-langserve
r/src/project-manager.ts' },
     position: { line: 0, character: 9 } } }

```

alexsaveliev:
Closes #141

### DIFF
--- a/src/connection.ts
+++ b/src/connection.ts
@@ -3,8 +3,8 @@ import { EventEmitter } from 'events';
 import { camelCase, omit } from 'lodash';
 import { FORMAT_TEXT_MAP, Span, Tracer } from 'opentracing';
 import { inspect } from 'util';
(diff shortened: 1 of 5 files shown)


felixfbecker:
@alexsaveliev moved logging responsibility to MessageWriter and MessageEmitter so we don't have multiple places where we have to log messages.

PR #145, opened 2017-04-07
Protecting TypeScript library files

alexsaveliev:
- excluding them from regular files and putting to a separate object. This is done to prevent attempts to re-fetch them (by build server for example). This may lead to unwanted promise rejection and lack of hover/definition information. Addresses sourcegraph/sourcegraph#5192


src/project-manager.ts
```diff
@@ -729,6 +730,25 @@ export class InMemoryFileSystem implements ts.ParseConfigHost, ts.ModuleResoluti
...
+	}
+
+	isTypeScriptLibrary(path: string): boolean {
+		return this.typeScriptLibraries.has(util.toUnixPath(path));
+	}
+
+	/**
+	 * Collects TypeScript library files
+	 */
+	private initTypeScriptLibraries() {
```
felixfbecker: This is doing sync IO when the InMemoryFileSystem is created, which is done for every new connection and will then block anything happening in other connections. Before it was done on startup, which was better.

src/project-manager.ts
```diff
@@ -698,6 +694,11 @@ export interface FileSystemNode {
 export class InMemoryFileSystem implements ts.ParseConfigHost, ts.ModuleResolutionHost {
 
 	/**
+	 * Contains in-memory map (path => content) for all TypeScript libraries
+	 */
+	private typeScriptLibraries = new Map<string, string>();
```
felixfbecker: Idea: could we maybe just save the typescript libraries in the normal `files` Map, but with a custom URI scheme like `tslib:`? That's how I did it in PHP: https://sourcegraph.com/github.com/felixfbecker/php-language-server/-/blob/src/ComposerScripts.php#L55:1-56:1

felixfbecker:
Should also have a test (can be a unit test)

src/project-manager.ts
alexsaveliev: I think that keeping them in a separate object is safer :smile:

src/memfs.ts
```diff
@@ -0,0 +1,288 @@
...
+
+/**
+ * TypeScript library files fetched from the local file system (bundled TS)
+ */
+let tsLibraries: Map<string, string>;
+
+/**
+ * Fetches TypeScript library files from local file system
+ */
+export function getTypeScriptLibraries(): Map<string, string> {
```
felixfbecker: Instead of this getter function, lets just use the map directly. Does it even have to be exported?

src/memfs.ts
```diff
@@ -0,0 +1,288 @@
...
+	 */
+	rootNode: FileSystemNode;
+
+	constructor(path: string, private logger: Logger = new NoopLogger()) {
+		this.path = path;
+		this.overlay = new Map<string, string>();
+		this.rootNode = { file: false, children: new Map<string, FileSystemNode>() };
+	}
+
+	isTypeScriptLibrary(path: string): boolean {
```
felixfbecker: I don't see a reason for this to be a method of InMemorySystem now

alexsaveliev: fixed

src/memfs.ts
alexsaveliev: exposing map directly. It's still needs to be exported for tests

PR #151, opened 2017-04-09
Ensure files refactor

### LINKED ISSUE #119: Consolidate ensureFilesFor*

felixfbecker:
Closes #119 

Method | Change
--------|-------------------------------
`ensureFilesForWorkspaceSymbol` | renamed to `ensureOwnFiles`
`ensureFilesForReferences` | removed, call `ensureAllFiles` directly
`ensureFilesForHoverAndDefinition` | removed, call`ensureFileReferences` directly
`ensureTransitiveFileDependencies` | renamed to `ensureFileReferences`, move logic to `resolveFileReferences`, return Observable
`ensureModuleStructure` | doesn't call `refreshFileTree` anymore
`refreshFileTree` | removed

Also
- Make these methods all take real URIs, not file paths or "relative" URIs.
- Trace them all properly
- Provide public API for invalidation of memoization caches, `invalidateModuleStructure` clears all the other caches too.
- Make ProjectConfiguration API sync, they weren't doing anything async.


src/project-manager.ts
```diff
@@ -326,64 +254,124 @@ export class ProjectManager implements Disposable {
...
+									this.rootPath,
+									resolver.dirname(filePath),
+									util.toUnixPath(referencedFile.fileName)
+								)
+							)
+						))
+				);
+			})
+			// Use same scheme, slashes, host for referenced URI as input file
+			.map(filePath => url.format({ ...parts, pathname: filePath.split(/[\\\/]/).map(encodeURIComponent).join('/'), search: undefined, hash: undefined }))
```
alexsaveliev: I'v seen few `split - encode - join`. Let's introduce a method maybe?

felixfbecker: it's a bit different here because I only want to encode the pathname component, not the whole URI, because everything else is taken from the input URI. In general you are right

Issue #99, opened 2017-03-14
Remove global util.setStrict()


felixfbecker:
If any of the utility functions need to behave differently, they should receive it as a parameter.

Especially for pure functions like `path2uri()` I highly doubt that they should behave differently at all for strict mode or the host OS: https://sourcegraph.com/github.com/sourcegraph/javascript-typescript-langserver/-/blob/src/util.ts#L72:1-73:1

Since this is called in tests, AppVeyor tests don't actually test any different behavior on Windows.

PR #156, opened 2017-04-11
workspace/symbol corrections

alexsaveliev:
These changes are targeted to resolve missing j2d from sourcegraph/javascript-typescript-langserver to microsoft/TypeScript definition

- corrected package extraction (it wasn't worked properly on local file system), added basic project manager functionality tests (for local and VFS modes) to cover fix with unit tests
- Handling compilation errors and internal TS errors: if there was a compilation error program object cannot be built. We should tolerate this case and provide no data
- **SIGNIFICANT**: Restricting files for default configuration object, giving access to them only if there were no other objects found. Otherwise it may collect files that weren't supposed to be compiled

### DIFF
--- a/src/project-manager.ts
+++ b/src/project-manager.ts
@@ -771,10 +771,7 @@ export class ProjectConfiguration {
 	 * Note that it does not perform any parsing or typechecking
 	 * @return program object (cached result of parsing and typechecking done by TS service)
 	 */
-	getProgram(): ts.Program {
-		if (!this.program) {
-			throw new Error('project is uninitialized');
-		}
+	getProgram(): ts.Program | undefined {
 		return this.program;
 	}
 
@@ -794,7 +791,11 @@ export class ProjectConfiguration {
 	 * the whole project or in some files
 	 */
 	syncProgram(): void {
-		this.program = this.getService().getProgram();
+		try {
+			this.program = this.getService().getProgram();
+		} catch (e) {
+			this.logger.error(`Cannot create program object for ${this.rootFilePath}`, e);
+		}
 	}
 
 	private initialized = false;
@@ -852,7 +853,7 @@ export class ProjectConfiguration {
 			this.logger
 		);
 		this.service = ts.createLanguageService(this.host, ts.createDocumentRegistry());
-		this.program = this.service.getProgram();
+		this.syncProgram();
 		this.initialized = true;
 	}
 
@@ -874,10 +875,16 @@ export class ProjectConfiguration {
 		}
 
 		this.init();
+
+		const program = this.getProgram();
+		if (!program) {
+			return;
+		}
+
 		let changed = false;
 		for (const fileName of (this.getHost().expectedFilePaths || [])) {
 			if (util.isGlobalTSFile(fileName) || (!util.isDependencyFile(fileName) && util.isDeclarationFile(fileName))) {
-				const sourceFile = this.getProgram().getSourceFile(fileName);
+				const sourceFile = program.getSourceFile(fileName);
 				if (!sourceFile) {
 					this.getHost().addFile(fileName);
 					changed = true;
@@ -886,7 +893,7 @@ export class ProjectConfiguration {
 		}
 		if (changed) {
 			// requery program object to synchonize LanguageService's data
-			this.program = this.getService().getProgram();
+			this.syncProgram();
 		}
 		this.ensuredBasicFiles = true;
 	}
@@ -900,14 +907,17 @@ export class ProjectConfiguration {
 		if (this.ensuredAllFiles) {
 			return;
 		}
-
 		this.init();
 		if (this.getHost().complete) {
 			return;
 		}
+		const program = this.getProgram();
+		if (!program) {
+			return;
+		}
 		let changed = false;
 		for (const fileName of (this.getHost().expectedFilePaths || [])) {
-			const sourceFile = this.getProgram().getSourceFile(fileName);
+			const sourceFile = program.getSourceFile(fileName);
 			if (!sourceFile) {
 				this.getHost().addFile(fileName);
 				changed = true;
@@ -915,7 +925,7 @@ export class ProjectConfiguration {
 		}
 		if (changed) {
 			// requery program object to synchonize LanguageService's data
-			this.program = this.getService().getProgram();
+			this.syncProgram();
 		}
 		this.getHost().complete = true;
 		this.ensuredAllFiles = true;
--- a/src/project-manager.ts
+++ b/src/project-manager.ts
@@ -483,13 +483,18 @@ export class ProjectManager implements Disposable {
 		}
 		if (!rootdirs.has('') && !this.configs.has('')) {
 			// collecting all the files in workspace by making fake configuration object
-			this.configs.set('', new ProjectConfiguration(this.localFs, '/', this.versions, '', {
+			const tsConfig: any = {
 				compilerOptions: {
 					module: ts.ModuleKind.CommonJS,
 					allowNonTsExtensions: false,
 					allowJs: true
 				}
-			}, this.traceModuleResolution, this.logger));
+			};
+			// if there is at least one config, giving no files to default one
+			if (this.configs.size > 0) {
+				tsConfig.exclude = ['**/*'];
+			}			
+			this.configs.set('', new ProjectConfiguration(this.localFs, '/', this.versions, '', tsConfig, this.traceModuleResolution, this.logger));
 		}
 	}
 }
--- a/src/project-manager.ts
+++ b/src/project-manager.ts
@@ -491,6 +491,7 @@ export class ProjectManager implements Disposable {
 				}
 			};
 			// if there is at least one config, giving no files to default one
+			// TODO: it makes impossible to IntelliSense gulpfile.js if there is a tsconfig.json in subdirectory
 			if (this.configs.size > 0) {
 				tsConfig.exclude = ['**/*'];
 			}
--- a/src/project-manager.ts
+++ b/src/project-manager.ts
@@ -493,7 +493,7 @@ export class ProjectManager implements Disposable {
 			// if there is at least one config, giving no files to default one
 			if (this.configs.size > 0) {
 				tsConfig.exclude = ['**/*'];
-			}			
+			}
 			this.configs.set('', new ProjectConfiguration(this.localFs, '/', this.versions, '', tsConfig, this.traceModuleResolution, this.logger));
 		}
 	}
--- a/src/project-manager.ts
+++ b/src/project-manager.ts
@@ -702,7 +702,7 @@ export class ProjectConfiguration {
 
 	/**
 	 * @param fs file system to use
-	 * @param rootFilePath root file path, relative to workspace hierarchy root
+	 * @param rootFilePath root file path, absolute
 	 * @param configFilePath configuration file path (relative to workspace root)
 	 * @param configContent optional configuration content to use instead of reading configuration file)
 	 */
@@ -749,8 +749,8 @@ export class ProjectConfiguration {
 			if (this.fs.fileExists(pkgJsonFile)) {
 				return JSON.parse(this.fs.readFile(pkgJsonFile)).name;
 			}
-			const parentDir = path_.dirname(currentDir);
-			if (parentDir === '.' || parentDir === '/' || parentDir === currentDir) {
+			const parentDir = path_.posix.dirname(currentDir);
+			if (parentDir === '.' || parentDir === currentDir || currentDir === this.fs.path) {
 				return undefined;
 			}
 			currentDir = parentDir;
--- a/src/test/project-manager.test.ts
+++ b/src/test/project-manager.test.ts
@@ -12,7 +12,7 @@ describe('ProjectManager', () => {
 	let projectManager: ProjectManager;
 
 	describe('getPackageName()', () => {
-		before(() => {
+		before(async () => {
 			const memfs = new InMemoryFileSystem('/');
 			const localfs = new MapFileSystem(new Map([
 				['file:///package.json', '{"name": "package-name-1"}'],
@@ -22,7 +22,7 @@ describe('ProjectManager', () => {
 			]));
 			const updater = new FileSystemUpdater(localfs, memfs);
 			projectManager = new ProjectManager('/', memfs, updater, true);
-			projectManager.ensureAllFiles();
+			await projectManager.ensureAllFiles();
 		});
 		it('should resolve package name when package.json is at the same level', () => {
 			assert.equal(projectManager.getConfiguration('').getPackageName(), 'package-name-1');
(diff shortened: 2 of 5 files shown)


src/fs.ts
```diff
@@ -78,6 +78,33 @@ export class LocalFileSystem implements FileSystem {
 	}
 }
 
+export class MapFileSystem implements FileSystem {
```
felixfbecker: I would put this in a file under the test folder

src/fs.ts
```diff
@@ -78,6 +78,33 @@ export class LocalFileSystem implements FileSystem {
...
+	 */
+	protected resolveUriToPath(uri: string): string {
+		return uri2path(uri).substring(1);
+	}
+
+	async getWorkspaceFiles(base?: string): Promise<Iterable<string>> {
+		const prefix = base ? this.resolveUriToPath(base) : '';
+		return iterate(this.files.keys())
+			.filter(path => !prefix || path.startsWith(prefix))
+			.map(path => 'file:///' + path.split('/').map(encodeURIComponent).join('/'));
```
felixfbecker: why save file paths when you convert to URIs here?

src/fs.ts
```diff
@@ -78,6 +78,33 @@ export class LocalFileSystem implements FileSystem {
...
 
+export class MapFileSystem implements FileSystem {
+
+	constructor(private files: Map<string, string>) {}
+
+	/**
+	 * Converts the URI to an absolute path
+	 */
+	protected resolveUriToPath(uri: string): string {
+		return uri2path(uri).substring(1);
```
felixfbecker: Two different URIs like `file:///1` and `inmemory:/1` would override each other and return the same content

src/project-manager.ts
```diff
@@ -479,19 +479,24 @@ export class ProjectManager implements Disposable {
...
 			if (!this.configs.has(dir)) {
-				this.configs.set(dir, new ProjectConfiguration(this.localFs, path_.posix.join('/', dir), this.versions, relativeFilePath, undefined, this.traceModuleResolution, this.logger));
+				this.configs.set(dir, new ProjectConfiguration(this.localFs, path_.posix.join(this.rootPath, dir), this.versions, relativeFilePath, undefined, this.traceModuleResolution, this.logger));
 			}
 			rootdirs.add(dir);
 		}
 		if (!rootdirs.has('') && !this.configs.has('')) {
 			// collecting all the files in workspace by making fake configuration object
-			this.configs.set('', new ProjectConfiguration(this.localFs, '/', this.versions, '', {
+			const tsConfig: any = {
```
felixfbecker: I would be surprised if TS didn't have a type for this

src/project-manager.ts
```diff
@@ -796,7 +798,11 @@ export class ProjectConfiguration {
 	 * the whole project or in some files
 	 */
 	syncProgram(): void {
-		this.program = this.getService().getProgram();
+		try {
+			this.program = this.getService().getProgram();
+		} catch (e) {
+			this.logger.error(`Cannot create program object for ${this.rootFilePath}`, e);
```
felixfbecker: Should code just continue on in this case without reacting to the error?

src/test/project-manager-helpers.ts
```diff
@@ -0,0 +1,35 @@
...
+	['subdirectory-with-tsconfig/src/tsconfig.json', '{}'],
+	['subdirectory-with-tsconfig/src/dummy.ts', '']
+];
+
+/**
+ * Describe a ProjectManager class
+ *
+ * @param createService Factory function to create the TypeScriptService instance to describe
+ */
+export async function describeProjectManager(factory: ProjectManagerFactory) {
```
felixfbecker: I don't think we need this, we only need to support subclassing `TypeScriptService`

src/test/project-manager.local.test.ts
```diff
@@ -0,0 +1,44 @@
...
+		memFileSystem = new InMemoryFileSystem(temporaryDir);
+		updater = new FileSystemUpdater(fileSystem, memFileSystem);
+	});
+	after(async () => {
+		await new Promise<void>((resolve, reject) => {
+			rimraf(temporaryDir, err => err ? reject(err) : resolve());
+		});
+	});
+	describeProjectManager(() => {
+		setStrict(false);
```
felixfbecker: I would prefer to solve #99

src/test/project-manager.local.test.ts
```diff
@@ -0,0 +1,44 @@
...
+import * as path from 'path';
+import * as rimraf from 'rimraf';
+import * as temp from 'temp';
+import { FileSystemUpdater, LocalFileSystem } from '../fs';
+import { InMemoryFileSystem } from '../memfs';
+import { ProjectManager } from '../project-manager';
+import { setStrict, toUnixPath } from '../util';
+import { describeProjectManager, expectedFiles } from './project-manager-helpers';
+
+describe('Local ProjectManager', () => {
```
felixfbecker: A unit test should only test the unit under test and should assume all dependencies (like the passed FileSystem interface) behave correctly. We shouldn't test every possible combination, or we get into an n x n problem. There should be one unit test suite for ProjectManager that tests it behaves correctly given a correctly-behaving file system (like the MapFileSystem) and one suite for e.g. LocalFileSystem that tests it behaves correctly too.

src/test/typescript-service-helpers.ts
```diff
@@ -75,18 +81,18 @@ export function describeTypeScriptService(createService: TypeScriptServiceFactor
 	describe('Workspace without project files', function (this: TestContext) {
 
 		beforeEach(initializeTypeScriptService(createService, new Map([
-			['file:///a.ts', 'const abc = 1; console.log(abc);'],
-			['file:///foo/b.ts', [
+			['a.ts', 'const abc = 1; console.log(abc);'],
+			['foo/b.ts', [
```
felixfbecker: Please revert this, I changed this on purpose. The LS shouldn't assume files are always file:// URIs, this will just make it harder to test non-file URIs in the future. The test itself also shouldn't rely on helper functions like `path2Uri`

src/test/typescript-service-helpers.ts
```diff
@@ -29,28 +26,37 @@ export interface TestContext {
...
+};
+
+/**
+ * Returns a function that initializes the test context with a TypeScriptService instance and initializes it (to be used in `beforeEach`)
+ *
+ * @param createService A factory that creates the TypeScript service. Allows to test subclasses of TypeScriptService
+ * @param fs file system that will serve files
+ * @param rootPath root path to be sent in `initialize` request
+ */
+export const initializeTypeScriptServiceWithFileSystem = (createService: TypeScriptServiceFactory, fs: FileSystem, rootPath: string) => async function (this: TestContext): Promise<void> {
```
felixfbecker: What was the motivation for making rootPath configurable?

src/typescript-service.ts
```diff
@@ -1535,7 +1544,11 @@ export class TypeScriptService {
 	 * @param fileName file name to fetch source file for or create it
 	 */
 	private _getSourceFile(configuration: pm.ProjectConfiguration, fileName: string): ts.SourceFile | null {
```
felixfbecker: Could this return `ts.SourceFile | undefined` to stay consistent?

src/typescript-service.ts
```diff
@@ -1544,8 +1557,9 @@ export class TypeScriptService {
 		}
 		configuration.getHost().addFile(fileName);
 		configuration.syncProgram();
-
-		return configuration.getProgram().getSourceFile(fileName);
+		// update program object
+		program = configuration.getProgram();
+		return program ? program.getSourceFile(fileName) : null;
```
felixfbecker: ... that would simplify this to `return program && program.getSourceFile(fileName)`

src/typescript-service.ts
```diff
@@ -1590,33 +1604,41 @@ export class TypeScriptService {
...
+						const items = config.getService().getNavigateToItems(symQuery.name || '', undefined, undefined, false);
+						const packageName = config.getPackageName();
+						const pd = packageName ? { name: packageName } : undefined;
+						for (const item of items) {
+							const sd = SymbolDescriptor.create(item.kind, item.name, item.containerKind, item.containerName, pd);
+							if (!util.symbolDescriptorMatch(symQuery, sd)) {
+								continue;
+							}
+							const si = this._transformNavItem(this.root, program, item);
+							if (!util.isLocalUri(si.location.uri)) {
```
felixfbecker: The behavior of this function is not correct, it checks the scheme not the host. What is this check for?
(I realised this is just reindented, but noticed anyway)

src/project-manager.ts
alexsaveliev: Well, surprise then! :smile:

src/project-manager.ts
alexsaveliev: I think so. We got a broken code, why should we die because of other sins?  :smile:

felixfbecker: Not die, but may fail the operation. I was worried that if the error is swallowed, the calling function will just continue and return incorrect results or error otherwise, hiding the true error reason.

src/typescript-service.ts
```diff
@@ -325,6 +325,10 @@ export class TypeScriptService {
 				const configuration = this.projectManager.getConfiguration(fileName);
 				// Ensure all files have been added
 				configuration.ensureAllFiles();
+				const program = configuration.getProgram();
+				if (!program) {
+					return Observable.from([]);
```
felixfbecker: you can just return `[]`, `mergeMap` implicitly calls `Observable.from()` on the return value

src/project-manager.ts
```diff
@@ -483,13 +483,18 @@ export class ProjectManager implements Disposable {
...
-			this.configs.set('', new ProjectConfiguration(this.localFs, '/', this.versions, '', {
+			const tsConfig: any = {
 				compilerOptions: {
 					module: ts.ModuleKind.CommonJS,
 					allowNonTsExtensions: false,
 					allowJs: true
 				}
-			}, this.traceModuleResolution, this.logger));
+			};
+			// if there is at least one config, giving no files to default one
```
felixfbecker: Do we even need the default config then? Or what is its purpose if it doesn't contain any files?

felixfbecker:
Would also love to get the new tests in, but can do that in a different PR :)

src/project-manager.ts
alexsaveliev: My idea here is:
- if there are no config files as all, let's make a default 'catch-them-all' config
- if there are some config files, let's make default one to handle situation when information is requested from a file that doesn't belong to any config (we should stay alive) but make it empty to exclude files which werent't expected to be compilable

src/typescript-service.ts
alexsaveliev: ok

src/project-manager.ts
felixfbecker: Agree with 1) and agree that we should stay alive (we shouldn't worry about "staying alive" == process crash, we have solid error handling at the RPC layer), but does the empty configuration in the second case actually provide any features? Or would it be equivalent to just bailing out because no configuration was found for the file?

What about the case where you have a tsconfig in e.g. `./src/tsconfig.json`, but also a gulpfile.js at the root? Can we still provide IntelliSense for the gulpfile.js?

alexsaveliev: It's complicated IMHO. We may have a situation when default config file lists everything in the project which may be not a good idea (as seen with microsoft/TypeScript repo). 
With my changes - no, we won't be able to IntelliSense gulpfile.

alexsaveliev:
added unit test

src/project-manager.ts
felixfbecker: Okay, I think we can accept that as a known limitation. But it would be good if you could mention this here in a TODO comment

alexsaveliev: aye, aye, captain, TODO added

src/test/project-manager.test.ts
```diff
@@ -0,0 +1,29 @@
...
+import chaiAsPromised = require('chai-as-promised');
+chai.use(chaiAsPromised);
+const assert = chai.assert;
+import { TypeScriptService } from '../typescript-service';
+import { initializeTypeScriptService, shutdownTypeScriptService, TestContext } from './typescript-service-helpers';
+
+describe('ProjectManager', () => {
+	describe('getPackageName()', function (this: TestContext) {
+		before(async function (this: TestContext) {
+			await initializeTypeScriptService(
```
felixfbecker: I would prefer if this just created a ProjectManager with a mocks. Might require to move the FS mock into an extra file.

alexsaveliev: OK, here is my attempt to create just a `ProjectManager` :smile:

src/test/project-manager.test.ts
```diff
@@ -0,0 +1,37 @@
+import * as chai from 'chai';
+import chaiAsPromised = require('chai-as-promised');
+chai.use(chaiAsPromised);
+const assert = chai.assert;
+import { FileSystemUpdater } from '../fs';
+import { InMemoryFileSystem } from '../memfs';
+import { ProjectManager } from '../project-manager';
+import { MapFileSystem } from './fs-helpers';
+
+function initialize(files: Map<string, string>): ProjectManager {
```
felixfbecker: No need for an external function

src/fs.ts
```diff
@@ -99,7 +99,7 @@ export class FileSystemUpdater {
 	 */
 	private concurrencyLimit = new Semaphore(100);
 
-	constructor(private remoteFs: FileSystem, private inMemoryFs: InMemoryFileSystem) {}
+	constructor(public remoteFs: FileSystem, public inMemoryFs: InMemoryFileSystem) {}
```
felixfbecker: I don't like this change. FileSystemUpdater has a dependency on `InMemoryFileSystem` and `ProjectManager` has a dependency on `InMemoryFileSystem`. By making them public and not explicitly passing them to `ProjectManager` you make the dependencies unclear. For example, ProjectManager now has access to the remote file system, but should only interact with the Updater.

src/test/project-manager.test.ts
alexsaveliev: done

src/fs.ts
alexsaveliev: reverted

src/test/project-manager.test.ts
```diff
@@ -0,0 +1,34 @@
...
+			const memfs = new InMemoryFileSystem('/');
+			const localfs = new MapFileSystem(new Map([
+				['file:///package.json', '{"name": "package-name-1"}'],
+				['file:///subdirectory-with-tsconfig/package.json', '{"name": "package-name-2"}'],
+				['file:///subdirectory-with-tsconfig/src/tsconfig.json', '{}'],
+				['file:///subdirectory-with-tsconfig/src/dummy.ts', '']
+			]));
+			const updater = new FileSystemUpdater(localfs, memfs);
+			projectManager = new ProjectManager('/', memfs, updater, true);
+			projectManager.ensureAllFiles();
```
felixfbecker: This is an async function, but it's not awaited

Issue #146, opened 2017-04-08
No test for references


felixfbecker:
`textDocument/references` is not covered at all. This is a core feature and needs a test.
I almost broke it in a refactor because all tests were passing.

https://codecov.io/gh/sourcegraph/javascript-typescript-langserver/src/master/src/typescript-service.ts#L265

PR #161, opened 2017-04-12
Add tests for textDocument/references

tomv564:
Tests empty, one in-file and multiple imported references

For #146

### DIFF
--- a/src/test/typescript-service-helpers.ts
+++ b/src/test/typescript-service-helpers.ts
@@ -1578,6 +1578,34 @@ export function describeTypeScriptService(createService: TypeScriptServiceFactor
 			assert.deepEqual(result, []);
 		} as any);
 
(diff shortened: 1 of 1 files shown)


src/test/typescript-service-helpers.ts
```diff
@@ -1531,6 +1531,122 @@ export function describeTypeScriptService(createService: TypeScriptServiceFactor
...
+		it('should provide an empty response when no reference is found', async function (this: TestContext) {
+			const result = await this.service.textDocumentReferences({
+				textDocument: {
+					uri: 'file:///a.ts'
+				},
+				position: {
+					line: 0,
+					character: 0
+				},
+				context: { includeDeclaration: false }
```
felixfbecker: We should probably test the behaviour of `includeDeclaration`

felixfbecker:
Thanks a lot!

PR #166, opened 2017-04-13
Filter out references from node_modules

felixfbecker:
As a user, I only want to know references in my own project, not the thousands of potential references in node_modules.

@alexsaveliev is this the only way to do it or is there a way to not let TS search node_modules in the first place?


alexsaveliev:
@felixfbecker I know no TS-specific way of doing it

PR #167, opened 2017-04-13
Trace syncProgram()

### LINKED ISSUE #147: Trace compilation
Tracing is currently only added to IO operations. We need to trace how long compilation takes. I am not sure which method actually does this. @alexsaveliev?

felixfbecker:
Closes #147 

Please check if I missed anything.

### DIFF
--- a/src/project-manager.ts
+++ b/src/project-manager.ts
@@ -267,7 +267,7 @@ export class ProjectManager implements Disposable {
 		ignore.add(uri);
 		return Observable.from(this.ensureModuleStructure(span))
 			// If max depth was reached, don't go any further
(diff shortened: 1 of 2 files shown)


felixfbecker:
@alexsaveliev notifications are not traced

alexsaveliev:
I am unsure about the following change
```
-	didClose(filePath: string) {
 +	didClose(filePath: string, span = new Span()) {
```
but these methods still called w/o span object. So what's the reason to introduce `span` argument? :smile:

felixfbecker:
So they _can_ be called with a span, even if they are not atm. A method shouldn't be coupled to its consumer ;)

PR #172, opened 2017-04-14
Fixed URI parsing on Windows

alexsaveliev:


src/project-manager.ts
```diff
@@ -359,7 +354,7 @@ export class ProjectManager implements Disposable {
 				);
 			})
 			// Use same scheme, slashes, host for referenced URI as input file
-			.map(filePath => url.format({ ...parts, pathname: filePath.split(/[\\\/]/).map(encodeURIComponent).join('/'), search: undefined, hash: undefined }))
+			.map(filePath => util.path2uri('', filePath))
```
felixfbecker: This changes the behaviour I noted in the comment above, which is using the same schems, slashes and host. Instead this now becomes a `file://` URI without host, even if the input was something different. I am trying to decouple the LS from this implication that everything is working with localhost `file://` URIs, this is a step backwards.

src/typescript-service.ts
```diff
@@ -439,11 +438,8 @@ export class TypeScriptService {
 	}
 
 	protected async _workspaceSymbolDefinitelyTyped(params: WorkspaceSymbolParams): Promise<SymbolInformation[] | null> {
-		const rootUriParts = url.parse(this.rootUri);
-		if (!rootUriParts.pathname) {
-			return null;
-		}
-		const packageJsonUri = url.format({ ...rootUriParts, pathname: path_.posix.join(rootUriParts.pathname, 'package.json') });
+		const normRootUri = this.rootUri.endsWith('/') ? this.rootUri : this.rootUri + '/';
+		const packageJsonUri = normRootUri + 'package.json';
```
felixfbecker: I don't understand why this change is needed, `path.join()` handles trailing slashes:
```js
> require('path').posix.join('hello/', 'world')
'hello/world'
> require('path').posix.join('hello', 'world')
'hello/world'
> require('path').posix.join('hello', '/world')
'hello/world'
> require('path').posix.join('hello/', '/world')
'hello/world'
```

this change doesn't work if the root URI has a query or hash component

src/project-manager.ts
```diff
@@ -316,12 +315,8 @@ export class ProjectManager implements Disposable {
 		if (observable) {
 			return observable;
 		}
-		const parts = url.parse(uri);
-		if (!parts.pathname) {
-			return Observable.throw(new Error(`Invalid URI ${uri}`));
-		}
 		// TypeScript works with file paths, not URIs
-		const filePath = parts.pathname.split('/').map(decodeURIComponent).join('/');
+		const filePath = util.uri2path(uri);
```
felixfbecker: I don't see anything incorrect about the now deleted URI-to-path conversion. What is the difference here, except that the version before handled non-file URIs and query, hash components? Does the TS service require paths to be in the format of the host OS?

src/typescript-service.ts
alexsaveliev: `path.join()` doesn't handle URL concatenation

```
path.posix.join('file:///foo', 'bar')
```
`file:/foo/bar`

felixfbecker: But it's not called on the URI here, it's called on the `pathname` of the URI

felixfbecker:
```
┌─────────────────────────────────────────────────────────────────────────────┐
│                                    href                                     │
├──────────┬┬───────────┬─────────────────┬───────────────────────────┬───────┤
│ protocol ││   auth    │      host       │           path            │ hash  │
│          ││           ├──────────┬──────┼──────────┬────────────────┤       │
│          ││           │ hostname │ port │ pathname │     search     │       │
│          ││           │          │      │          ├─┬──────────────┤       │
│          ││           │          │      │          │ │    query     │       │
"  http:   // user:pass @ host.com : 8080   /p/a/t/h  ?  query=string   #hash "
│          ││           │          │      │          │ │              │       │
└──────────┴┴───────────┴──────────┴──────┴──────────┴─┴──────────────┴───────┘
```

alexsaveliev:
The reason of my change is that pathname of Windows-style file URL is `/C:/foo/bar` (note the leading slash) so we shouldn't mix `url.parse()` and hand-made `uri2path`, `path2uri`

felixfbecker:
Sorry, still don't get it - Unix pathnames start with a slash too

src/project-manager.ts
alexsaveliev: No, TS service does not require path to be in format of the host OS but we shouldn't use url.parse() in some places while leaving uri2path in others, let's be consistent. I'm OK to use url.parse() if we will be able to access local files after translation (you won't be able to access `/c:/foo/bar` on Windows)

src/project-manager.ts
alexsaveliev: OK, maybe this is a step backward. What if you'll change uri2path and path2uri to use url.parse() and reverse function?

felixfbecker: That would work, but it would be less efficient here because we already have the URI parsed, so we would parse it twice. What we could do is accept `string | url.Url`

alexsaveliev:
> Sorry, still don't get it - Unix pathnames start with a slash too

They are, but you can't access content of /c:/foo/bar on Windows :cry:

src/project-manager.ts
felixfbecker: But to me, it is always unclear what these functions do exactly. For example, does `uri2path` always return a Unix path? If no, does the path style depend on the host or client OS? Do we have use cases for both? How to differentiate? That's why until now I mostly did the conversion "manually"

src/project-manager.ts
felixfbecker: Oh, right! makes sense. I think we should match the path for a drive letter + colon (colon is reserved on Unix as a PATH seperator) and strip the leading slash in that case.

felixfbecker:
Would we be able to test these failures if we fixed #99?

alexsaveliev:
> Would we be able to test these failures if we fixed #99?

Not sure that fixing #99 alone would help.
I'm trying to keep in mind that potentially we should be able to work with combinations of (vfs, local fs) and (windows, unix) that's why I've introduced `strict` flag. If we want to get rid of it we should somehow distinguish VFS situation when (potentially) file:///c:/foo/bar refers to some resource /c:/foo/bar and local FS situation when file:///c:/foo/bar refers to c:\foo\bar. Not sure that idea to check if pathname starts with letter followed by colon would work in case of abstract URI where colon may not mark logical drive

felixfbecker:
On a file:// URI the only place where a colon could be present in the pathname is behind a windows drive letter. It's the only way to detect the client OS. I used that technique in multiple projects:

- https://sourcegraph.com/github.com/felixfbecker/vscode-php-debug/-/blob/src/paths.ts#L15:1-16:1
- https://sourcegraph.com/github.com/felixfbecker/php-language-server/-/blob/src/utils.php#L46:1-47:1

But I am wondering why we need that. Why do we need to "access" files of a specific path? There is no guarantee we can access them on disk anyway, so the only way to access them is our `FileSystem` interface, which works with URIs. The only component in our app that needs to work with true file paths is TS, and TS shoudn't care about what kind of paths it gets as it never reads them from disk and has to deal with different kinds of paths anyway.

> If we want to get rid of it we should somehow distinguish VFS situation when (potentially) file:///c:/foo/bar refers to some resource /c:/foo/bar and local FS situation when file:///c:/foo/bar refers to c:\foo\bar.

Why do we need to distinguish? The URI is the resource identifier, not the path. The URI allows us to retrieve the file content from our `FileSystem` interface no matter whether we work with `ClientFileSystem` or `LocalFileSystem`.

src/typescript-service.ts
felixfbecker: agree. c:/foo/bar is completely wrong though imo. It's unix path style, but doesn't begin with a slash, which makes it a relative path.

PR #173, opened 2017-04-15
Improve dispatcher

felixfbecker:
- handle sync errors
- log result
- unsubscribe on close, error, exit to not leak subscriptions

### DIFF
--- a/src/connection.ts
+++ b/src/connection.ts
@@ -173,6 +173,9 @@ export function registerLanguageHandler(messageEmitter: MessageEmitter, messageW
 			case 'exit':
 				// Ignore exit notification, it's not the responsibility of the TypeScriptService to handle it,
 				// but the TCP / STDIO server which needs to close the socket or kill the process
+				for (const subscription of subscriptions.values()) {
+					subscription.unsubscribe();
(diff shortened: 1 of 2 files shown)


src/connection.ts
```diff
@@ -173,6 +173,9 @@ export function registerLanguageHandler(messageEmitter: MessageEmitter, messageW
 			case 'exit':
 				// Ignore exit notification, it's not the responsibility of the TypeScriptService to handle it,
 				// but the TCP / STDIO server which needs to close the socket or kill the process
+				for (const subscription of subscriptions.values()) {
```
alexsaveliev: why not `iterate` / `forEach` ? :wink:

felixfbecker: Good question. This code is for side effects. `iterate`/Observables are great where you have a return value and need to filter/map/etc. But `forEach` has many disadvantages:
- you can't step through the loop with a debugger
- no `break` keyword
- TS compiler can't narrow types of properties because the callback could be called after the object was mutated

so in general, prefer `map`/`filter` over `for..of`, but `for..of` over `forEach`.

Issue #170, opened 2017-04-14
Completions / hover never uses updated content after initial didOpen


tomv564:
Repro:
1. Open a TS file (triggers didOpen) 
2. Hover over a variable, observe the correct information is returned
3. Rename the variable (triggers didChange with incrementing version number  1, 2, 3, ...)
4. Hover again: Observe the original name of the variable is returned.
5. Save the document (triggers didSave)
6. Hover again: Observe the original name is still returned

Only when the server is restarted does the hover return the new name.

I tried this (originally with completions) on a few different commits, it stopped working where the ensureFiles refactor was merged (https://github.com/sourcegraph/javascript-typescript-langserver/pull/151)

felixfbecker:
Thanks for filing, could you provide an LSP message log (see `--trace`, `--logfile`) so we can see what URIs your client sent? Thanks!

tomv564:
Log attached here: [lspserver.txt](https://github.com/sourcegraph/javascript-typescript-langserver/files/923049/lspserver.txt)

alexsaveliev:
claimed fixed by #175

PR #175, opened 2017-04-18
Corrected content changes tracking

alexsaveliev:
- using absolute URI to track content changes (didOpen/didClose/didSave/didChange), to read overlay content and to set/get script snapshot versions
- addresses #170


src/memfs.ts
```diff
@@ -141,22 +139,16 @@ export class InMemoryFileSystem implements ts.ParseConfigHost, ts.ModuleResoluti
...
 	private readFileIfExists(path: string): string | undefined {
-
-		let content = this.overlay.get(path);
-		if (content !== undefined) {
-			return content;
-		}
-
-		const rel = path_.posix.relative('/', path);
-		content = this.overlay.get(rel);
+		const uri = util.path2uri(this.path, path);
```
felixfbecker: We should pass `rootUri` to the constructor as we cannot guarantee a `file://` URI

alexsaveliev:
@felixfbecker addressed your comment, PTAL

src/memfs.ts
```diff
@@ -233,6 +232,11 @@ export class InMemoryFileSystem implements ts.ParseConfigHost, ts.ModuleResoluti
 	trace(message: string) {
 		this.logger.log(message);
 	}
+
+	// TODO: move this method to util and use through all the code
+	private path2uri(path: string): string {
```
felixfbecker: I see it resolves relative paths, in what cases is it called with a relative path?

alexsaveliev: Here are few attempts to read files from memfs, see the mix of abs and relative paths:

```
? tsconfig.json
? C:/projects/javascript-typescript-langserver/node_modules/typescript/lib/lib.d
.ts
? d:/tmp/javascript-typescript-langserver/src/a.ts
```

felixfbecker: Is that only in `readFile()`? I see a problem here because `util.resolve()` only looks at strict mode or host OS, to it may not detect a relative path properly. There is also `url.resolve()`, which may be helpful here.

alexsaveliev: new method is used in `readFile()` and `fileExists()`. `util.resolve()` indeed looks at strict mode (expected POSIX-style resolution) and host OS (On Windows we expect to use Windows-specific resolution unless we are working in strict mode). `url.resolve()` probably may fail if OS-specific path is passed

tomv564:
I quickly tested this branch and confirmed it fixed the issue I was having 👍

felixfbecker:
I am merging this in without the last commit, it turned out to be more complex then I thought and I don't want to block this fix.

PR #177, opened 2017-04-18
Removed relative path resolution for referenced files

alexsaveliev:
- do not resolving referenced files relative to root anymore because they are converted directly into URI w/o additional resolution back to abs path

@tomv564 will appreciate unit test, thanks!


tomv564:
Awesome fixes today, will try to contribute with unit tests for all reference types!

PR #178, opened 2017-04-18
Filter out workspace symbols from node_modules

felixfbecker:
workspace/symbol should only search the own workspace if I am not mistaken.

### DIFF
--- a/src/typescript-service.ts
+++ b/src/typescript-service.ts
@@ -429,6 +429,7 @@ export class TypeScriptService {
 		const symbols = iterate(configs)
 			.map(config => this._collectWorkspaceSymbols(config, query || symbolQuery, limit))
 			.flatten<SymbolInformation>()
+			.filter(symbol => !symbol.location.uri.includes('/node_modules/'))
 			.take(limit)
 			.toArray();
 		// Save empty query result
--- a/src/typescript-service.ts
+++ b/src/typescript-service.ts
@@ -848,22 +848,6 @@ export class TypeScriptService {
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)


src/typescript-service.ts
```diff
@@ -429,6 +429,7 @@ export class TypeScriptService {
 		const symbols = iterate(configs)
 			.map(config => this._collectWorkspaceSymbols(config, query || symbolQuery, limit))
 			.flatten<SymbolInformation>()
+			.filter(symbol => !symbol.location.uri.includes('/node_modules/'))
```
alexsaveliev: We may get less results than expected because of the following code above:
```
.map(config => this._collectWorkspaceSymbols(config, query || symbolQuery, limit))
```

felixfbecker: Sorry, not following? You mean if you pass a symbol query that is defined in a dependency?

felixfbecker: https://github.com/sourcegraph/sourcegraph/issues/5396

alexsaveliev: > You mean if you pass a symbol query that is defined in a dependency?

No, I mean that `this._collectWorkspaceSymbols` takes `limit` so it probably returns up to N symbols but then they might be filtered out by the added `filter` call so there might be less results

felixfbecker: yeah, that is true. But I don't see a way around that.

alexsaveliev: ~prepare for trouble~ make it double in a call of _collect..?

beyang:
Add a test case for this behavior.

PR #180, opened 2017-04-18
Do not add global .d.ts files in node_modules

felixfbecker:
This caused definitions in node_modules, like `node_modules/@reactivex/rxjs/typings/globals/node/node.d.ts` to be used instead of `node_modules/@types/node/index.d.ts`.
If a global file is supposed to be included, it should be included explicitly through tsconfig.json or a triple slash reference.

Also removes the last usage of `walkInMemoryFs` (we have an iterator API for that now)

### DIFF
--- a/src/memfs.ts
+++ b/src/memfs.ts
@@ -236,33 +236,6 @@ export class InMemoryFileSystem implements ts.ParseConfigHost, ts.ModuleResoluti
 }
 
 /**
- * Iterates over in-memory cache calling given function on each node until callback signals abort or all nodes were traversed
- */
-export function walkInMemoryFs(fs: InMemoryFileSystem, rootdir: string, walkfn: (path: string, isdir: boolean) => Error | void): Error | void {
-	const err = walkfn(rootdir, true);
-	if (err) {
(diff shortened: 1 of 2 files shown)


beyang:
Please add a test case for this behavior.

felixfbecker:
Will do as a follow-up

tomv564:
I'm not sure if this statement should apply to the `@types` folder in `node_modules`:

> If a global file is supposed to be included, it should be included explicitly through tsconfig.json or a triple slash reference.

The language service now complains about `cannot find name 'define'` or `'it'` in this project's test files, while the mocha declarations are already installed. Vanilla VS Code has no issues showing hover info for `define` or `it`.

The TypeScript handbook suggests `npm install --save @types/lodash` should be all that is needed: http://www.typescriptlang.org/docs/handbook/declaration-files/consumption.html
Also discussed in the Angular docs at https://angular.io/docs/ts/latest/guide/typescript-configuration.html#!#installable-typings-files

felixfbecker:
We could special-case `@types` packages

PR #181, opened 2017-04-18
Add test for ensureReferencedFiles and implement types reference

### LINKED ISSUE #169: Fails to resolve triple-slash reference
Starting up the service against this codebase reproduces the issue, I am calling initialize with:

```
initialize {'rootUri': 'file:///Users/tomv/Projects/sourcegraph/javascript-typescript-langserver'} 
```

When processing the triple-slash reference to thenable.d.ts in `vscode-jsonrpc/lib/main.d.ts`:

ProjectManager converts this reference to a relative path (`node_modules/vscode-jsonrpc/lib/thenable.d.ts`) at https://github.com/sourcegraph/javascript-typescript-langserver/blob/master/src/project-manager.ts#L350,

which is later converted to an the URI `file://node_modules/vscode-jsonrpc/lib/thenable.d.ts` at https://github.com/sourcegraph/javascript-typescript-langserver/blob/master/src/project-manager.ts#L362

The following error is logged by the process:
```
Error resolving file references for file:///Users/tomv/Projects/sourcegraph/javascript-typescript-langserver/node_modules/vscode-jsonrpc/lib/main.d.ts: { Error: ENOENT: no such file or directory, open '/node_modules/vscode-jsonrpc/lib/thenable.d.ts'
  errno: -2,
  code: 'ENOENT',
  syscall: 'open',
  path: '/node_modules/vscode-jsonrpc/lib/thenable.d.ts' } 
```

tomv564:
Contains support for this type of directive:
```typescript
/// <reference types="node"/>
```
And a test for **project-manager** for normal `import` statements and this type of directive:
```typescript
/// <reference path="./mytypes.d.ts"/>
```
Should close #169.


src/test/project-manager.test.ts
```diff
@@ -31,4 +32,24 @@ describe('ProjectManager', () => {
...
 		});
 	});
+	describe('ensureReferencedFiles', () => {
+		before(async () => {
+			memfs = new InMemoryFileSystem('/');
+			const localfs = new MapFileSystem(new Map([
+				['file:///package.json', '{"name": "package-name-1"}'],
+				['file:///node_modules/somelib/index.js', '/// <reference path="./pathref.d.ts"/>\n/// <reference types="node"/>'],
+				['file:///node_modules/somelib/pathref.d.ts', ''],
+				['file:///node_modules/@types/node/index.d.ts', ''],
```
felixfbecker: %-encode `@`

src/test/project-manager.test.ts
```diff
@@ -31,4 +32,24 @@ describe('ProjectManager', () => {
...
+				['file:///src/dummy.ts', 'import * as somelib from "somelib";']
+			]));
+			const updater = new FileSystemUpdater(localfs, memfs);
+			projectManager = new ProjectManager('/', memfs, updater, true);
+			await projectManager.ensureReferencedFiles('file:///src/dummy.ts').toPromise();
```
felixfbecker: This is the unit under test, it should be called in the test, not the setup code.

src/test/project-manager.test.ts
```diff
@@ -31,4 +32,24 @@ describe('ProjectManager', () => {
...
+		});
+		it('should resolve import reference', () => {
```
felixfbecker: s/resolve/ensure

src/test/project-manager.test.ts
```diff
@@ -31,4 +32,24 @@ describe('ProjectManager', () => {
...
+			memfs.getContent('file:///node_modules/somelib/index.js');
```
felixfbecker: these are not awaited, so if they throw an exception it won't actually fail the test.

tomv564: I did not `await` the getContent calls but the test does fail if getContent throws an exception.

felixfbecker: My mistake. These functions are synchronous

PR #182, opened 2017-04-18
Correct DefinitelyTyped workspace/symbol handling

felixfbecker:
- DefinitelyTyped changed their structure so that all packages are now in the `types/` subdirectory
- workspace/symbol on DefinitelyTyped was incredibly slow


beyang:
@felixfbecker looks good to me, but can you please add a test case for this case?

felixfbecker:
Well, there are already tests for this, but they asserted functionality with the old DT structure. I corrected them. We can't really test performance without having to clone the whole DT repo, which I wanna avoid.

PR #183, opened 2017-04-20
Add support for multiple configurations

### LINKED ISSUE #124: Remove file://PATH URIs
`file://PATH` is an invalid URL. If a URL starts with a double slash and does not have a host, the path must start with a slash. There is no thing as a relative `file://` URL.

alexsaveliev:
- there might be a chance that multiple configurations are present (jsconfig + tsconfig), keeping them both, otherwise they may be mutually exclusive and may prevent files collection
- removed unused code
- using absolute paths when searching for file's configuration and for keeping files in memory FS
- addresses references to https://github.com/Microsoft/vscode/-/blob/src/vs/platform/telemetry/common/telemetryService.ts#L103:2-103:11

Closes #124


src/project-manager.ts
```diff
@@ -11,6 +11,10 @@ import { Logger, NoopLogger } from './logging';
 import { InMemoryFileSystem } from './memfs';
 import * as util from './util';
 
+const typeTypeScript = 'ts';
+const typeJavaScript = 'js';
+const types = [typeTypeScript, typeJavaScript];
```
felixfbecker: Instead of defining constants, I would define `type ConfigType = 'jsconfig' | 'tsconfig'`

src/test/project-manager.test.ts
```diff
@@ -52,4 +52,24 @@ describe('ProjectManager', () => {
...
+			]));
+			const updater = new FileSystemUpdater(localfs, memfs);
+			projectManager = new ProjectManager('/', memfs, updater, true);
+			await projectManager.ensureAllFiles();
+		});
+		it('should resolve best configuration based on file name', () => {
+			const jsConfig = projectManager.getConfiguration('/src/foo.js');
+			const tsConfig = projectManager.getConfiguration('/src/foo.ts');
+			assert.isNotNull(jsConfig);
+			assert.isNotNull(tsConfig);
```
felixfbecker: this is not possible with the return type of `getConfiguration()`

src/typescript-service.ts
```diff
@@ -497,7 +489,8 @@ export class TypeScriptService {
 			span.log({ event: 'fetched package files' });
 
 			// Search symbol in configuration
-			const config = this.projectManager.getConfiguration(packageRoot);
+			// forcing TypeScript mode
+			const config = this.projectManager.getConfiguration(packageRoot + 'file.ts');
```
felixfbecker: this feels hacky. I would add an optional parameter `type` to `getConfiguration()` that specifies the type and defaults to `this.getConfigurationType(fileName)`

src/util.ts
```diff
@@ -107,22 +107,6 @@ export function uri2path(uri: string): string {
...
-	root = toUnixPath(root);
-	if (uri.startsWith(root)) {
-		uri = uri.substring(root.length);
-	}
-	while (uri.startsWith('/')) {
-		uri = uri.substring(1);
-	}
-	return uri;
-}
-
```
felixfbecker: feels good

src/project-manager.ts
alexsaveliev: Changed to `enum`. Okay?

src/test/project-manager.test.ts
alexsaveliev: Removed null tests

src/project-manager.ts
```diff
@@ -472,39 +482,67 @@ export class ProjectManager implements Disposable {
...
+			if (pos <= 0) {
 				dir = '';
+			} else {
+				dir = dir.substring(0, pos);
+			}
+			const configType = this.getConfigurationType(filePath);
+			let configs = this.configs.get(configType);
+			if (!configs) {
+				configs = new Map<string, ProjectConfiguration>();
+				this.configs.set(configType, configs);
```
felixfbecker: We only have two fixed types of configurations. We should initialize these maps in the constructor. It might be even easier (and faster) to have simply two attributes, `jsConfigs` and `tsConfigs`.

alexsaveliev: as you wish

src/project-manager.ts
```diff
@@ -472,39 +488,73 @@ export class ProjectManager implements Disposable {
...
+					allowJs: false
+				}
+			};
+			tsConfig.include = ['**/*.ts', '**/*.tsx'];
+			// if there is at least one config, giving no files to default one
+			// TODO: it makes impossible to IntelliSense gulpfile.js if there is a tsconfig.json in subdirectory
+			if (this.tsConfigs.size > 0) {
+				tsConfig.exclude = ['**/*'];
+			}
+			this.tsConfigs.set('', new ProjectConfiguration(this.localFs, this.rootPath, this.versions, '', tsConfig, this.traceModuleResolution, this.logger));
```
felixfbecker: This is not DRY. Should be possible to do
```ts
for (const configs of [this.tsConfigs, this.jsConfigs]) {
  // ...
}
```

src/project-manager.ts
```diff
@@ -472,39 +488,73 @@ export class ProjectManager implements Disposable {
...
+		}
+	}
+
+	private getConfigurationType(filePath: string): ConfigType {
```
felixfbecker: add docblock

src/project-manager.ts
```diff
@@ -472,39 +488,73 @@ export class ProjectManager implements Disposable {
...
+		const name = path_.posix.basename(filePath);
+		if (name === 'tsconfig.json') {
+			return ConfigType.TsConfig;
+		} else if (name === 'jsconfig.json') {
```
felixfbecker: `else` is not needed

src/project-manager.ts
```diff
@@ -26,11 +28,14 @@ export class ProjectManager implements Disposable {
...
 	/**
-	 * Workspace subtree (folder) -> JS/TS configuration mapping.
+	 * (Workspace subtree (folder) -> TS or JS configuration) mapping.
 	 * Configuration settings for a source file A are located in the closest parent folder of A.
 	 * Map keys are relative (to workspace root) paths
 	 */
-	private configs: Map<string, ProjectConfiguration>;
+	private configs = {
+		js: new Map<string, ProjectConfiguration>(),
+		ts: new Map<string, ProjectConfiguration>()
```
felixfbecker: I just thought, why do we wanna keep different maps at all? The only reason they conflict is because we decided to map from directory name to ProjectConfiguration. Why not simply map from jsconfig.json/tsconfig.json path (or URI) to ProjectConfiguration?

Issue #162, opened 2017-04-12
Implement diagnostics


tomv564:
For the users who use the server through their editor, project-wide diagnostics is probably pretty high on the wish list.

I hacked around a bit in [diagnostics-hack](/tomv564/javascript-typescript-langserver/tree/diagnostics-hack), but this functionality probably needs careful design. 

Do you have any further requirements (eg. should be optional?) and ideas (eg. how to convert didChange events into eventual notifications)? 
Is there a minimal scope for an initial implementation?

felixfbecker:
We definitely need this. It should definitely not poll with a setTimeout, but publish them when they become available. I am not familiar enough with the TS API to know what is provided here.

tomv564:
VS Code also only offers diagnostics for files currently open in the editor and suggests users to run another background task (`tsc -w`) to get project-wide errors on save.

There is discussion and work being done to have a running tsserver instance respect the **compileOnSave** tsconfig.json setting at https://github.com/Microsoft/vscode/issues/7015

I believe the simplest implementation would behave like VS Code (only open files). Do you feel the language service should also publish project diagnostics at startup and save?

felixfbecker:
@tomv564 that is not true, you can report diagnostics at any time. For example, the PHP language server reports them as soon as they are detected during indexing.

tomv564:
Agreed, I was exploring why the existing TS support in VS Code *doesn't* send diagnostics for all files all the time. I'll try to get the same behaviour as the PHP language server if the TS API allows for it.

tomv564:
Great, diagnostics are flowing once the first file is opened. Two considerations:
* Problems with ill-configured typings in dependencies create a really distracting amount of diagnostics. 
  * Should we apply some kind of strategy to prevent huge payloads of diagnostics from being re-published (when the file in question never changes)?
* The TS service exposes diagnostics for some common tsconfig errors (eg. No files matched include pattern). We could expose these as well!

Let me know if you would like to close this issue, I can create separate issues for the above.

felixfbecker:
Not republishing not-changed diagnostics sounds good. I'm wondering whether there is an efficient way in the compiler API to do this vs deep array comparison.

It also looks like diagnostics had a perf impact, a lot more tests are marked as slow now:
before: https://travis-ci.org/sourcegraph/javascript-typescript-langserver/builds/227149983
after: https://travis-ci.org/sourcegraph/javascript-typescript-langserver/builds/227189384

[...]

PR #186, opened 2017-04-20
Diagnostics for opened files

tomv564:
This is a small implementation towards #162:
* ProjectManager exposes files to compiler via ProjectConfiguration in didOpen/didChange
* ProjectManager passes a DiagnosticsHandler to ProjectConfiguration
* ~~Expose `programSyncEvents` on ProjectConfiguration, triggered after compilation~~
* ~~DiagnosticsPublisher requests diagnostics on a synced program and publishes per file~~
* ProjectConfiguration passes TS file diagnostics to DiagnosticsHandler 
* ~~Basic tests~~
  * Unit tests for DiagnosticsPublisher
  * Existing functional tests expanded to test correct uri and content update handling

> Uncertainties:
> * It feels totally reasonable to expose a file to the compiler on didOpen, but for some reason this was not being done yet?
> * As discussed, VS Code stops showing diagnostics on broken code when you close the file. We would have to remove the file from the compiler to achieve consistent behaviour with VS Code.
> * Missing test scenarios for broken tsconfig, missing node modules, etc.
> * A minimum delay of 1 second feels good on instantly-compiling projects, but perhaps it should be the editor's job to delay showing feedback while the user is typing? Or should the didChange handler not trigger compiles immediately?

Thanks for the answers on these! Continuing conversation below.

### DIFF
--- a/src/diagnostics.ts
+++ b/src/diagnostics.ts
@@ -38,11 +38,11 @@ export class DiagnosticsPublisher implements DiagnosticsHandler {
 		const diagnosticsByFile = this.groupByFile(diagnostics);
 
 		// add empty diagnostics for fixed files, so client marks them as resolved
-		this.problemFiles.forEach(file => {
+		for (const file of this.problemFiles) {
 			if (!diagnosticsByFile.has(file)) {
 				diagnosticsByFile.set(file, []);
 			}
-		});
+		}
 		this.problemFiles.clear();
 
 		// for each file: publish and set as problem file
@@ -101,14 +101,14 @@ export class DiagnosticsPublisher implements DiagnosticsHandler {
 	 */
 	private groupByFile(diagnostics: ts.Diagnostic[]): Map<string, ts.Diagnostic[]> {
 		const diagnosticsByFile: Map<string, ts.Diagnostic[]> = new Map();
-		diagnostics.forEach(d => {
-			const diagnosticsForFile = diagnosticsByFile.get(d.file.fileName);
+		for (const diagnostic of diagnostics) {
+			const diagnosticsForFile = diagnosticsByFile.get(diagnostic.file.fileName);
 			if (!diagnosticsForFile) {
-				diagnosticsByFile.set(d.file.fileName, [d]);
+				diagnosticsByFile.set(diagnostic.file.fileName, [diagnostic]);
 			} else {
-				diagnosticsForFile.push(d);
+				diagnosticsForFile.push(diagnostic);
 			}
-		});
+		}
 		return diagnosticsByFile;
 	}
 }
--- a/src/diagnostics.ts
+++ b/src/diagnostics.ts
@@ -100,7 +100,7 @@ export class DiagnosticsPublisher implements DiagnosticsHandler {
 	 * @param diagnostics All diagnostics received in an update
 	 */
 	private groupByFile(diagnostics: ts.Diagnostic[]): Map<string, ts.Diagnostic[]> {
-		const diagnosticsByFile: Map<string, ts.Diagnostic[]> = new Map();
+		const diagnosticsByFile = new Map<string, ts.Diagnostic[]>();
 		for (const diagnostic of diagnostics) {
 			const diagnosticsForFile = diagnosticsByFile.get(diagnostic.file.fileName);
 			if (!diagnosticsForFile) {
--- a/src/diagnostics.ts
+++ b/src/diagnostics.ts
@@ -1,0 +1,114 @@
+import * as ts from 'typescript';
+import { DiagnosticSeverity, PublishDiagnosticsParams } from 'vscode-languageserver';
+import { LanguageClient } from './lang-handler';
+import * as util from './util';
+
+/**
+ * Receives file diagnostics (typically implemented to send diagnostics to client)
+ */
+export interface DiagnosticsHandler {
+	updateFileDiagnostics(diagnostics: ts.Diagnostic[]): void;
+}
+
+/**
+ * Forwards diagnostics from typescript calls to LSP diagnostics
+ */
+export class DiagnosticsPublisher implements DiagnosticsHandler {
+	/**
+	 * The files that were last reported to have errors
+	 * If they don't appear in the next update, we must publish empty diagnostics for them.
+	 */
+	private problemFiles: Set<string> = new Set();
+
+	/**
+	 * Requires a connection to the remote client to send diagnostics to
+	 * @param remoteClient
+	 */
+	constructor(private remoteClient: LanguageClient) {}
+
+	/**
+	 * Receives file diagnostics from eg. ts.getPreEmitDiagnostics
+	 * Diagnostics are grouped and published by file, empty diagnostics are sent for files
+	 * not present in subsequent updates.
+	 * @param diagnostics
+	 */
+	updateFileDiagnostics(diagnostics: ts.Diagnostic[]): void {
+
+		// categorize diagnostics by file
+		const diagnosticsByFile = this.groupByFile(diagnostics);
+
+		// add empty diagnostics for fixed files, so client marks them as resolved
+		this.problemFiles.forEach(file => {
+			if (!diagnosticsByFile.has(file)) {
+				diagnosticsByFile.set(file, []);
+			}
+		});
+		this.problemFiles.clear();
+
+		// for each file: publish and set as problem file
+		diagnosticsByFile.forEach((diagnostics, file) => {
+			this.publishFileDiagnostics(file, diagnostics);
+			if (diagnostics.length > 0) {
+				this.problemFiles.add(file);
+			}
+		});
+	}
+
+	/**
+	 * Converts a diagnostic category to an LSP DiagnosticSeverity
+	 * @param category The Typescript DiagnosticCategory
+	 */
+	private parseDiagnosticCategory(category: ts.DiagnosticCategory): DiagnosticSeverity {
+		switch (category) {
+			case ts.DiagnosticCategory.Error:
+				return DiagnosticSeverity.Error;
+			case ts.DiagnosticCategory.Warning:
+				return DiagnosticSeverity.Warning;
+			case ts.DiagnosticCategory.Message:
+				return DiagnosticSeverity.Information;
+				// unmapped: DiagnosticSeverity.Hint
+		}
+	}
+
+	/**
+	 * Sends given diagnostics for a file to the remote client
+	 * @param file Absolute path as specified from the TS API
+	 * @param diagnostics Matching file diagnostics from the TS API, empty to clear errors for file
+	 */
+	private publishFileDiagnostics(file: string, diagnostics: ts.Diagnostic[]): void {
+		const params: PublishDiagnosticsParams = {
+			uri: util.path2uri('', file),
+			diagnostics: diagnostics.map(d => {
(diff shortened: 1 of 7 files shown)


felixfbecker:
It is quite hard for me to understand what is going on here (some documentaiton on DiagnosticsPublisher would be helpful). What I got:

- ProjectConfiguration now emits an event with a new project whenever it is updated
- The ProjectManager subscribes to this event, but it does so by being passed to ProjectManager, which passes it to ProjectConfiguration, which then passes an event Subject to a method.
- The DiagnosticsPublisher takes a Subscribable to subscribe on and maintains a map of file to collected diagnostics. On a program update, it updates the map, then publishes them all.

Problems I have:
- The point of evented API is that the event producer is not coupled to (doesn't know about) the event consumer. The producer just emits the event, the consumer subscribes to it.
- The `subscribe` method looks like the the one defined in the Rx `Subscribable` interface, but does the exact opposite. It doesn't take a Subscriber, it takes a Subscribable to subscribe on.
- Subjects are an anti pattern. If you cannot avoid them, their scope should be limited - publicly exposed should be only the Observable interface.
- I think using an Observable/Subject for the program update event is overkill. Are there any benefits over making the ProjectConfiguration and ProjectManager a simple EventEmitter?
- Tests: The unit test for ProjectManager should just test the unit under test (ProjectManager). That means it should just assert that the program update event is emitted.

> It feels totally reasonable to expose a file to the compiler on didOpen, but for some reason this was not being done yet?

No reason, makes sense to me. I think it didn't matter because we read all files in anyway and all operations ensured the files they needed.

> As discussed, VS Code stops showing diagnostics on broken code when you close the file. We would have to remove the file from the compiler to achieve consistent behaviour with VS Code.

I wouldn't be too worried about that, I don't have a problem with showing diagnostics for closed files.

> A minimum delay of 1 second feels good on instantly-compiling projects, but perhaps it should be the editor's job to delay showing feedback while the user is typing? Or should the didChange handler not trigger compiles immediately?

I would expect the client to not spam `didChange` on every key stroke, so I don't see that much of an issue with publishing diagnostics immediately.

felixfbecker:
Please ping when this is ready for another review

tomv564:
I've addressed points from the initial review, thanks for taking the time!

> It is quite hard for me to understand what is going on here
* Added docstrings, comments and broke up DiagnosticsPublisher's update function

> Using Rx is overkill
* Agreed, ProjectConfiguration now just calls `updateFileDiagnostics` on DiagnosticsPublisher

> The unit test for ProjectManager should just test the unit under test (ProjectManager)
* Created unit tests to test the details of DiagnosticsPublisher's work
* Kept the functional tests (titled 'Diagnostics') to catch regressions in the system

I have some time to work on this, but feel free to work on my branch!

tomv564:
Ping @felixfbecker, ready for another review!

src/project-manager.ts
```diff
@@ -951,6 +971,28 @@ export class ProjectConfiguration {
 	private ensuredAllFiles = false;
 
 	/**
+	 * Ensures a single file is available to the LanguageServiceHost
+	 * @param filePath
+	 */
+	ensureSourceFile(filePath: string): boolean {
```
felixfbecker: As I understand it, the return type indicates whether the program was resynced as part of the method call? Why exactly is that information needed for a caller who wants to ensure a source file? If it is needed to decide whether to republish diagnostics, then this puts unnecessary burden on the callers to always check the return value and act on it - emitting an event is better in that case

src/test/diagnostics.test.ts
```diff
@@ -0,0 +1,219 @@
...
+		message: 'Failure B',
+		range: { end: { character: 4, line: 0 }, start: { character: 0, line: 0 } },
+		severity: 1,
+		code: 33,
+		source: 'ts'
+	};
+
+	beforeEach(() => {
+		publishSpy = sinon.spy();
+		langClient = {
```
felixfbecker: Since you want to stub all and only spy on one I would use `sinon.createStubInstance(RemoteLanguageClient)`

src/test/diagnostics.test.ts
```diff
@@ -0,0 +1,219 @@
...
+		diagnosticsManager.updateFileDiagnostics([]);
+		sinon.assert.calledWithExactly(publishSpy, {
+			diagnostics: [],
+			uri: 'file:///file1.ts'
+		});
+	});
+});
+
+// Integration tests that include TS compilation.
+describe('Diagnostics', () => {
```
felixfbecker: Still not convinced we should have these here.

The expected behaviour of the ProjectManager is to call the DiagnosticHandler, so there should be tests in the ProjectManager suite that assert it is called. That's where all the setup/teardown code is located.

All tests are to prevent regressions

src/test/diagnostics.test.ts
tomv564: Awesome!

src/test/diagnostics.test.ts
tomv564: Right you are, these have been replaced with some unit tests on didChange/didOpen.

tomv564:
@felixfbecker thanks for the guidance, hope the didChange and test changes are what you intended!

src/project-manager.ts
```diff
@@ -464,9 +471,10 @@ export class ProjectManager implements Disposable {
 		if (!config) {
 			return;
 		}
-		config.ensureConfigFile(span);
+		config.ensureConfigFile();
+		config.ensureSourceFile(filePath);
 		config.getHost().incProjectVersion();
-		config.syncProgram(span);
+		config.syncProgram();
```
felixfbecker: why the removal of the span?

src/diagnostics.ts
```diff
@@ -0,0 +1,118 @@
...
+		// add empty diagnostics for fixed files, so client marks them as resolved
+		for (const file of this.problemFiles) {
+			if (!diagnosticsByFile.has(file)) {
+				diagnosticsByFile.set(file, []);
+			}
+		}
+		this.problemFiles.clear();
+
+		// for each file: publish and set as problem file
+		diagnosticsByFile.forEach((diagnostics, file) => {
```
felixfbecker: `for of`

src/test/diagnostics.test.ts
```diff
@@ -0,0 +1,84 @@
...
+import chaiAsPromised = require('chai-as-promised');
+chai.use(chaiAsPromised);
+import * as sinon from 'sinon';
+import * as ts from 'typescript';
+import { DiagnosticsPublisher } from '../diagnostics';
+import { LanguageClient, RemoteLanguageClient } from '../lang-handler';
+
+describe('DiagnosticsPublisher', () => {
+	let langClient: LanguageClient;
+	let diagnosticsManager;
```
felixfbecker: no type

src/test/diagnostics.test.ts
```diff
@@ -0,0 +1,84 @@
...
+	const createTSFileDiagnostic = (message: string, file: ts.SourceFile) => {
+		return {
```
felixfbecker: use arrow function without return or named function

src/test/diagnostics.test.ts
```diff
@@ -0,0 +1,84 @@
...
+		range: { end: { character: 4, line: 0 }, start: { character: 0, line: 0 } },
+		severity: 1,
+		code: 33,
+		source: 'ts'
+	};
+
+	beforeEach(() => {
+		publishSpy = sinon.spy();
+		langClient = sinon.createStubInstance(RemoteLanguageClient);
+		langClient.textDocumentPublishDiagnostics = publishSpy;
```
felixfbecker: all methods are stubs now, so no need to assign the spy here. Just type the langClient as something like
```ts
{ [K in keyof LanguageClient]: LanguageClient[K] & sinon.SinonStub }
// or
LanguageClient & { textDocumentPublishDiagnostics: Function & sinon.SinonSpy }

File: src/test/project-manager.test.ts [hunk 9]
```diff
@@ -70,4 +75,46 @@ describe('ProjectManager', () => {
...
+				['file:///src/dummy.ts', 'const num: number = "banana";']
+			]));
+			const updater = new FileSystemUpdater(localfs, memfs);
+			projectManager = new ProjectManager('/', memfs, updater, diagnosticsPublisher, true);
+			await projectManager.ensureAllFiles();
+		});
+		it('should compile opened file and return diagnostics', async () => {
+			projectManager.didOpen('file:///src/dummy.ts', 'const num: number = "banana";');
+			const lastArgs = diagnosticsSpy.lastCall.args[0];
+			assert.equal(lastArgs.length, 1);
```
felixfbecker: Pretty sure there is a sinon assert function for this. Should also first assert that the function is called at all?

src/typescript-service.ts
```diff
@@ -67,7 +68,6 @@ export type TypeScriptServiceFactory = (client: LanguageClient, options?: TypeSc
  * underscore.
  */
 export class TypeScriptService {
-
```
felixfbecker: revert

src/diagnostics.ts
```diff
@@ -0,0 +1,118 @@
...
+	 * The files that were last reported to have errors
+	 * If they don't appear in the next update, we must publish empty diagnostics for them.
+	 */
+	private problemFiles: Set<string> = new Set();
+
+	/**
+	 * Requires a connection to the remote client to send diagnostics to
+	 * @param remoteClient
+	 */
+	constructor(private remoteClient: LanguageClient) {}
```
felixfbecker: rename to `client`

src/diagnostics.ts
```diff
@@ -0,0 +1,118 @@
...
+
+/**
+ * Forwards diagnostics from typescript calls to LSP diagnostics
+ */
+export class DiagnosticsPublisher implements DiagnosticsHandler {
+	/**
+	 * The files that were last reported to have errors
+	 * If they don't appear in the next update, we must publish empty diagnostics for them.
+	 */
+	private problemFiles: Set<string> = new Set();
```
felixfbecker: prefer type inference

src/diagnostics.ts
```diff
@@ -0,0 +1,118 @@
...
+
+	/**
+	 * Groups all diagnostics per file they were reported on so they can be stored and sent in batches
+	 * @param diagnostics All diagnostics received in an update
+	 */
+	private groupByFile(diagnostics: ts.Diagnostic[]): Map<string, ts.Diagnostic[]> {
+		const diagnosticsByFile = new Map<string, ts.Diagnostic[]>();
+		for (const diagnostic of diagnostics) {
+			// TODO for some reason non-file diagnostics end up in here (where file is undefined)
+			const fileName = diagnostic.file ? diagnostic.file.fileName : '';
```
felixfbecker: `const fileName = diagnostic.file && diagnostic.file.fileName`

src/test/project-manager.test.ts
tomv564: Added an assert as suggested, but I couldn't find a way to assert on the last call only with sinon's asserts. They suggest using `sinon.assert.calledWith`, and we would need a `sinon.match` to avoid check all of TS's  diagnostics type. I tried this briefly but ran out of time, let me know if you want me to switch it to this syntax.

tomv564:
Thanks for the review, apologies for the unintended changes!

felixfbecker:
Thanks! Great work

Issue #190, opened 2017-04-20
Identifier appears twice in Quick Open search 


lindaxie:
Repro:
- go to https://sourcegraph.com/github.com/angular-ui/ui-router
- open Quick Open and search for #makecomponenttemplate 
- notice how it appears twice in the Quick Open search results https://cl.ly/1E28350j1L3x and both results take you to the same definition

felixfbecker:
@alexsaveliev still happening. Should get a test

alexsaveliev:
@felixfbecker this happens because there are **two** servers (javascript and typescript) that return the same data. I think that results should be filtered to leave only distinct items or only one server from TS/JS pair should be called.

felixfbecker:
This would only be solved by https://github.com/sourcegraph/sourcegraph/issues/5401 then, but I would require major changes I would like to defer to the JS GA.

@lindaxie how common was this issue? Would it be acceptable to leave this as is for now?

lindaxie:
@felixfbecker I've only seen it once or twice but I haven't spent a lot of time testing workspace symbol search so it could be common. However, it's doesn't affect the user's ability to user Sourcegraph, it just looks strange and can be confusing for the user (hence P2).

Yes, we can defer this issue until JS GA.

PR #192, opened 2017-04-21
excluding duplicate workspace symbols

alexsaveliev:
- addresses #190


felixfbecker:
I wonder what the underlying cause of this is. Somewhere the TS Language service is apparently producing the same symbol item twice.

Once workspace/symbol uses Observables this can be solved with the `distinct()` operator.

alexsaveliev:
> I wonder what the underlying cause of this is. 

there might be few configurations that contains the same file(s) (explicit) or files from a different configurations may refer to the same file (implicit)

felixfbecker:
Makes sense! Could you add that as a comment to the filter as the reason why we need to do this?

alexsaveliev:
> Makes sense! Could you add that as a comment to the filter as the reason why we need to do this?

done

PR #203, opened 2017-04-27
Represent URIs as WHATWG URLs

felixfbecker:
This uses the WHATWG `URL` objects everywhere a URI was previously saved, passed or returned as a string and replaces usages of `url.parse()` and `url.format()`.  This makes it
- easier to work with the URIs - access segments without explicitly parsing before
- more typesafe, it's impossible now to pass a file path to a function taking a URI

Also reimplements `path2uri` and `uri2path`.
`rootUri` is now passed into more places to be used as the `path2uri` root.

@alexsaveliev any Windows incompatibilities you see?


package.json
```diff
@@ -61,7 +62,7 @@
     "@types/lodash": "4.14.63",
     "@types/mocha": "2.2.40",
     "@types/mz": "0.0.31",
-    "@types/node": "7.0.14",
+    "@types/node": "6.0.70",
```
alexsaveliev: Why, mister Anderson?

src/fs.ts
```diff
@@ -34,45 +34,54 @@ export class RemoteFileSystem implements FileSystem {
...
+	constructor(protected rootUri: URL) {}
 
 	/**
-	 * Converts the URI to an absolute path
+	 * Returns the file path where a given URI should be located on disk
 	 */
-	protected resolveUriToPath(uri: string): string {
-		return toUnixPath(path.resolve(this.rootPath, uri2path(uri)));
+	protected resolveUriToPath(uri: URL): string {
+		return uri2path(uri);
```
alexsaveliev: I'd remove `resolveUriToPath` method then and replace it with direct `uri2path` invocation

src/fs.ts
```diff
@@ -34,45 +34,54 @@ export class RemoteFileSystem implements FileSystem {
...
+				// Don't return directories
+				nodir: true,
+				// Search directories recursively
+				matchBase: true,
+				// Return absolute file paths
+				absolute: true
+			} as any, (err, matches) => err ? reject(err) : resolve(matches));
 		});
-		return iterate(files).map(file => baseUri + file.split('/').map(encodeURIComponent).join('/'));
+		return iterate(files).map(file => new URL(file.split('/').map(encodeURIComponent).join('/'), base.href + '/'));
```
alexsaveliev: Does it handle special characters properly? I see the following example in https://url.spec.whatwg.org/#url:

```
var input = "https://example.org/💩",
    url = new URL(input)
url.pathname // "/%F0%9F%92%A9"
```

There input is non-encoded

src/memfs.ts
```diff
@@ -53,8 +58,8 @@ export class InMemoryFileSystem implements ts.ParseConfigHost, ts.ModuleResoluti
 	/**
 	 * Returns an IterableIterator for all URIs known to exist in the workspace (content loaded or not)
 	 */
-	uris(): IterableIterator<string> {
-		return this.files.keys();
+	uris(): IterableIterator<URL> {
+		return iterate(this.files.keys()).map(uri => new URL(uri));
```
alexsaveliev: Java does it better `map(URL::new)` :grinning:

src/project-manager.ts
```diff
@@ -315,17 +315,13 @@ export class ProjectManager implements Disposable {
...
 		if (observable) {
 			return observable;
 		}
-		const parts = url.parse(uri);
-		if (!parts.pathname) {
-			return Observable.throw(new Error(`Invalid URI ${uri}`));
-		}
 		// TypeScript works with file paths, not URIs
-		const filePath = parts.pathname.split('/').map(decodeURIComponent).join('/');
+		const filePath = uri.pathname.split('/').map(decodeURIComponent).join('/');
```
alexsaveliev: `uri2path` for consistency?

src/project-manager.ts
```diff
@@ -370,16 +366,16 @@ export class ProjectManager implements Disposable {
 				);
 			})
 			// Use same scheme, slashes, host for referenced URI as input file
-			.map(filePath => url.format({ ...parts, pathname: filePath.split(/[\\\/]/).map(encodeURIComponent).join('/'), search: undefined, hash: undefined }))
+			.map(filePath => new URL(filePath.split(/[\\\/]/).map(encodeURIComponent).join('/'), uri.href))
```
alexsaveliev: `path2uri` for consistency?

src/util.ts
```diff
@@ -79,32 +79,41 @@ export function convertStringtoSymbolKind(kind: string): SymbolKind {
...
-		p = file;
+		parts[0] = encodeURIComponent(parts[0]);
+	}
+	// Encode all other parts
+	for (let i = 1; i < parts.length; i++) {
+		parts[i] = encodeURIComponent(parts[i]);
 	}
-	p = toUnixPath(p).split('/').map(encodeURIComponent).join('/');
-	return ret + p;
+	const pathname = parts.join(isWindowsUri ? '\\' : '/');
```
alexsaveliev: Isn't `isWindowsUri` check redundant here?

src/util.ts
```diff
@@ -79,32 +79,41 @@ export function convertStringtoSymbolKind(kind: string): SymbolKind {
...
+/**
+ * Returns the path component of the passed URI as a file path.
+ * The OS style and seperator is determined by the presence of a Windows drive letter + colon in the URI.
+ * Does not check the URI protocol.
+ */
+export function uri2path(uri: URL): string {
+	// %-decode parts
+	let filePath = decodeURIComponent(uri.pathname);
+	// Strip the leading slash on Windows
+	const isWindowsUri = /^\/[a-z]:\//.test(filePath);
```
alexsaveliev: please make case-insensitive

src/fs.ts
```diff
@@ -34,45 +34,54 @@ export class RemoteFileSystem implements FileSystem {
 	 * The files request is sent from the server to the client to request a list of all files in the workspace or inside the directory of the base parameter, if given.
 	 * A language server can use the result to index files by filtering and doing a content request for each text document of interest.
 	 */
-	async getWorkspaceFiles(base?: string, childOf = new Span()): Promise<Iterable<string>> {
-		return iterate(await this.client.workspaceXfiles({ base }, childOf))
-			.map(textDocument => textDocument.uri);
+	async getWorkspaceFiles(base?: URL, childOf = new Span()): Promise<Iterable<URL>> {
+		return iterate(await this.client.workspaceXfiles({ base: base && base.href }, childOf))
+			.map(textDocument => new URL(textDocument.uri));
```
alexsaveliev: Could we somehow made it consistent - either `new URI` across the code or `path2uri`? 
there can be only one

felixfbecker: `path2uri` for converting file paths, `new URL` for converting a string-URL to a `URL` object

src/fs.ts
felixfbecker: Tried it (if you look at commit log) but it is useful for overriding it. Otherwise the globbing etc. needs to be completely reimplemented.

src/fs.ts
felixfbecker: Are you suggesting to leave out the `encodeURIComponent` call?

src/project-manager.ts
felixfbecker: Do we want Windows paths here? (Probably we do)

src/project-manager.ts
felixfbecker: Makes sense

src/util.ts
felixfbecker: What do you mean? We need to decide which separator to use

src/project-manager.ts
alexsaveliev: I think that having unified POSIX paths would be better for consistency

src/util.ts
alexsaveliev: I'd leave POSIX style

src/fs.ts
alexsaveliev: ok

package.json
felixfbecker: Because we the minimum Node version we can rely on is Node 6 (LTS). If we used Node 7, we would be able to use `require('url').URL` instead of the polyfill

src/fs.ts
felixfbecker: You are correct, encoding is done automatically for the path, but not all characters. Specifically, `@` characters are not encoded. I don't think this is a problem.

src/project-manager.ts
felixfbecker: I would want a Windows user to see `C:\Users\felix\myfile.ts` as a `containerName` for a class though, not `/C:/Users/felix/myfile.ts`

PR #212, opened 2017-04-30
Add tracing spans around retrieving and publishing diagnostics

tomv564:
Adds a sibling span to `getProgram()` for `ts.getPreEmitDiagnostics`.
A 3rd sibling is added when publishing file diagnostics.

I was unable to test this yet, but can do this later after looking at the diagnostics performance regressions.

### DIFF
--- a/src/diagnostics.ts
+++ b/src/diagnostics.ts
@@ -1,4 +1,4 @@
-import { Span } from 'opentracing/lib';
+import { Span } from 'opentracing';
 import * as ts from 'typescript';
 import { DiagnosticSeverity, PublishDiagnosticsParams } from 'vscode-languageserver';
 import { LanguageClient } from './lang-handler';
@@ -34,28 +34,33 @@ export class DiagnosticsPublisher implements DiagnosticsHandler {
 	 * @param diagnostics
 	 */
 	updateFileDiagnostics(diagnostics: ts.Diagnostic[], childOf = new Span()): void {
-		const span = childOf.tracer().startSpan('Publish diagnostics', { childOf });
+		const span = childOf.tracer().startSpan('Update file diagnostics', { childOf });
 
-		// categorize diagnostics by file
-		const diagnosticsByFile = this.groupByFile(diagnostics);
+		try {
+			// categorize diagnostics by file
+			const diagnosticsByFile = this.groupByFile(diagnostics);
 
-		// add empty diagnostics for fixed files, so client marks them as resolved
-		for (const file of this.problemFiles) {
-			if (!diagnosticsByFile.has(file)) {
-				diagnosticsByFile.set(file, []);
+			// add empty diagnostics for fixed files, so client marks them as resolved
+			for (const file of this.problemFiles) {
+				if (!diagnosticsByFile.has(file)) {
+					diagnosticsByFile.set(file, []);
+				}
 			}
-		}
-		this.problemFiles.clear();
+			this.problemFiles.clear();
 
-		// for each file: publish and set as problem file
-		for (const [file, diagnostics] of diagnosticsByFile) {
-			this.publishFileDiagnostics(file, diagnostics);
-			if (diagnostics.length > 0) {
-				this.problemFiles.add(file);
+			// for each file: publish and set as problem file
+			for (const [file, diagnostics] of diagnosticsByFile) {
+				this.publishFileDiagnostics(file, diagnostics);
+				if (diagnostics.length > 0) {
+					this.problemFiles.add(file);
+				}
 			}
+		} catch (err) {
+			span.setTag('error', true);
+			span.log({ 'event': 'error', 'error.object': err, 'message': err.message, 'stack': err.stack});
+		} finally {
+			span.finish();
 		}
-
-		span.finish();
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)


src/diagnostics.ts
```diff
@@ -32,7 +33,8 @@ export class DiagnosticsPublisher implements DiagnosticsHandler {
 	 * not present in subsequent updates.
 	 * @param diagnostics
 	 */
-	updateFileDiagnostics(diagnostics: ts.Diagnostic[]): void {
+	updateFileDiagnostics(diagnostics: ts.Diagnostic[], childOf = new Span()): void {
+		const span = childOf.tracer().startSpan('Publish diagnostics', { childOf });
 
 		// categorize diagnostics by file
 		const diagnosticsByFile = this.groupByFile(diagnostics);
```
felixfbecker: the operation needs to be wrapped in a try/catch/finally because spans must always be finished. You can find many examples of that in the project, it's always the same pattern

src/project-manager.ts
```diff
@@ -858,16 +858,36 @@ export class ProjectConfiguration {
 	 * the whole project or in some files
 	 */
 	syncProgram(childOf = new Span()): void {
-		const span = childOf.tracer().startSpan('Sync program', { childOf });
+		const syncProgramSpan = childOf.tracer().startSpan('Sync program', { childOf });
```
felixfbecker: there should always be only one span per function for readability. Operations that are worth to be traced should be extracted into a new function

src/project-manager.ts
```diff
@@ -858,16 +858,36 @@ export class ProjectConfiguration {
...
-			span.finish();
+			syncProgramSpan.finish();
+		}
+		if (this.program) {
+			this.updateDiagnostics(this.program, childOf);
+		}
+	}
+
+	updateDiagnostics(program: ts.Program, childOf = new Span()): void {
+		const getDiagnosticsSpan = childOf.tracer().startSpan('Get diagnostics', { childOf });
```
felixfbecker: single span, name it `span` and the operation name a humanized version of the method name

src/project-manager.ts
```diff
@@ -858,16 +858,36 @@ export class ProjectConfiguration {
...
+			getDiagnosticsSpan.log({ event: 'result', result: diagnostics.length });
+		} catch (err) {
+			getDiagnosticsSpan.setTag('error', true);
+			getDiagnosticsSpan.log({ 'event': 'error', 'error.object': err, 'message': err.message, 'stack': err.stack});
+			this.logger.error(`Cannot get diagnostics for program at ${this.rootFilePath}`, err);
+		} finally {
+			getDiagnosticsSpan.finish();
+		}
+		if (diagnostics !== undefined) {
+			this.diagnosticsHandler.updateFileDiagnostics(diagnostics, childOf);
```
felixfbecker: nothing should happen in the function anymore after the span was finished, everything should happen inside the `try` block

src/diagnostics.ts
```diff
@@ -1,3 +1,4 @@
+import { Span } from 'opentracing/lib';
```
felixfbecker: remove `/lib`

src/project-manager.ts
```diff
@@ -861,14 +861,34 @@ export class ProjectConfiguration {
...
 			span.finish();
 		}
+		if (this.program) {
+			this.updateDiagnostics(this.program, childOf);
+		}
+	}
+
+	updateDiagnostics(program: ts.Program, childOf = new Span()): void {
+		const span = childOf.tracer().startSpan('Update diagnostics', { childOf });
+		let diagnostics: ts.Diagnostic[] | undefined;
```
felixfbecker: define this as const inside the `try`

src/project-manager.ts
```diff
@@ -861,14 +861,34 @@ export class ProjectConfiguration {
...
 		} catch (err) {
 			span.setTag('error', true);
 			span.log({ 'event': 'error', 'error.object': err, 'message': err.message, 'stack': err.stack });
 			this.logger.error(`Cannot create program object for ${this.rootFilePath}`, err);
 		} finally {
 			span.finish();
 		}
+		if (this.program) {
+			this.updateDiagnostics(this.program, childOf);
+		}
```
felixfbecker: this should be inside the `try` ans pass the `span`, not `childOf`

src/project-manager.ts
```diff
@@ -861,14 +861,34 @@ export class ProjectConfiguration {
...
+	}
+
+	updateDiagnostics(program: ts.Program, childOf = new Span()): void {
+		const span = childOf.tracer().startSpan('Update diagnostics', { childOf });
+		let diagnostics: ts.Diagnostic[] | undefined;
+		try {
+			diagnostics = ts.getPreEmitDiagnostics(program);
+			span.log({ event: 'result', result: diagnostics.length });
+			if (diagnostics !== undefined) {
```
felixfbecker: if `diagnostics` can be `undefined`, the log above will throw.

src/project-manager.ts
```diff
@@ -871,6 +871,21 @@ export class ProjectConfiguration {
 		}
 	}
 
+	updateDiagnostics(program: ts.Program, childOf = new Span()): void {
```
felixfbecker: should this be `private`?

PR #215, opened 2017-05-02
Only request diagnostics for js/ts files outside node_modules

tomv564:
This removes the overhead of calling getSemanticDiagnostics on package / built-in declaration files after every compile (syncProgram).

* Javascript and Typescript files outside node_modules should still return diagnostics.
* Latency on tests and operations that require a compile should be drastically improved

### DIFF
--- a/src/project-manager.ts
+++ b/src/project-manager.ts
@@ -874,7 +874,7 @@ export class ProjectConfiguration {
 		try {
 			let diagnostics: ts.Diagnostic[] = [];
 			for (const file of program.getSourceFiles()) {
-				if (!file.fileName.includes('/node_modules/') && util.isJSTSFile(file.fileName)) {
+				if (!/[\/\\]node_modules[\/\\]/.test(file.fileName)) {
 					diagnostics = diagnostics.concat(ts.getPreEmitDiagnostics(program, file));
 				}
 			}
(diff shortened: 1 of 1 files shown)


src/project-manager.ts
```diff
@@ -872,9 +872,14 @@ export class ProjectConfiguration {
 	private updateDiagnostics(program: ts.Program, childOf = new Span()): void {
 		const span = childOf.tracer().startSpan('Update diagnostics', { childOf });
 		try {
-			const diagnostics = ts.getPreEmitDiagnostics(program);
-			span.log({ event: 'result', result: diagnostics.length });
+			let diagnostics: ts.Diagnostic[] = [];
+			for (const file of program.getSourceFiles()) {
+				if (!file.fileName.includes('/node_modules/') && util.isJSTSFile(file.fileName)) {
```
felixfbecker: does this work on Windows? I would call `toUnixPath()` before or use `/[\/\\]node_modules[\/\\]/.test()`

why the check for `.isJSTSFile()`? What other files could there be, a tsconfig.json?

tomv564:
This built 2 minutes faster than latest master, but I'm hoping to run this branch for a bit today to see what kind of diagnostics go missing.

felixfbecker:
Thanks for fixing this so quick!

src/project-manager.ts
tomv564: Sounds reasonable, I will submit some fixes after using this branch on Windows

tomv564: This worked for me on windows, but I applied your fixes nonetheless!

Issue #216, opened 2017-05-02
Remove syncProgram()?


felixfbecker:
It looks to me like `LanguageService.getProgram()` was designed to be fast when already compiled, and compile on demand. I have the feeling that calling `syncProgram()` everywhere is slow because we call `syncProgram()` way more often than `getProgram()`. When I removed it it didn't break any tests besides diagnostics. Diagnostics would have to be emitted on `didOpen()` instead.

PR #219, opened 2017-05-03
Remove syncProgram()

felixfbecker:
Idea outlined #216 

No tests broke besides diagnostics.
@tomv564 diagnostics would have to be only reported for opened files in `didOpen`, thoughts?

### DIFF
--- a/src/project-manager.ts
+++ b/src/project-manager.ts
@@ -315,7 +315,7 @@ export class ProjectManager implements Disposable {
 		ignore.add(uri);
 		return Observable.from(this.ensureModuleStructure(span))
 			// If max depth was reached, don't go any further
-			.mergeMap(() => maxDepth === 0 ? [] : this.resolveReferencedFiles(uri, span))
+			.mergeMap(() => maxDepth === 0 ? [] : this.resolveReferencedFiles(uri))
 			// Prevent cycles
 			.filter(referencedUri => !ignore.has(referencedUri))
 			// Call method recursively with one less dep level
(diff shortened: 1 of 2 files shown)


tomv564:
`didOpen` and `didChange` should invoke getProgram() in a way so that diagnostics are collected afterwards. Perhaps `projectVersion` can be of some help here, especially if it is also incremented by `didOpen`?

felixfbecker:
Wouldn't it be enough if `didChange` and `didOpen` simply get the diagnostics of the source file that was opened/changed?

tomv564:
Yes, that would be an elegant fix to the performance issues.

It is probably worthwhile to revisit project-wide diagnostics at some point!

felixfbecker:
@beyang would like to get a review from you on this since you have more context on why this was done the way it was before

beyang:
`syncProgram` was necessary before, because not all files were fetched on workspace initialization. Files were fetched on demand as needed (only those files which the user directly accessed or their transitive dependencies).

If now, we ensure all files have been fetched (and made available to the LanguageService) before `getProgram` is called, then `syncProgram` should no longer be necessary.

felixfbecker:
We still only fetch transitive dependencies for hover and definition. But what is the benefit of calling `getProgram()` and then storing that Program object in the ProjectConfiguration vs calling `getProgram()` directly only when needed?

beyang:
Doesn't seem to be any clear benefit. I'd have to guess that the old way was done to cache things. But if the `getProgram` is cheap enough and called in less places, that might no longer be necessary.

PR #226, opened 2017-05-04
Only publish diagnostics on didOpen/didChange

### LINKED ISSUE #223: Diagnostics for whole workspace have huge performance hit
@tomv564 Unfortunately #215 did not improve the situation a lot. See this trace (yellow is the critical path):

image

For now I will have to disable them, it would be great if you could do a PR to switch to only report diagnostics for open files on didOpen/didChange.

felixfbecker:
Closes #223 

This only reports diagnostics on didOpen/didChange like tsserver.
Reverts a lot of the initial implementation since keeping track of problematic files is not needed anymore, but I reused some logic and tests.

/cc @tomv564


tomv564:
Nice and simple :)

Issue #149, opened 2017-04-09
Evented API


felixfbecker:
Currently, every time an update is done (new file etc) we explicitly call functions that throw away current state and rebuild it.

- it's inefficient, because the update methods need to do change detection
- it's prone to errors, because a developer can easily forget to call the appropiate update method
- it results in much coupling, because every object that updates state depends on all other objects that need to act on state changes and call its update methods

Example:
- `ensuredModuleStructure` is set to `undefined`
- `ensureModuleStructure` calls `refreshFileTree`
- `refreshFileTree` fetches file structure again
- iterates over all of them to check what files to fetch
- calls `createConfigurations` which also iterates over all files to create new configurations
- finally resets all configurations, discarding compilation state
- next `ensureBasicFiles` call will iterate over all files again, adding them back

Instead, an object should emit _events_ for updates that require other objects to update, and these objects can _listen_ on that.

Example:
- file is added to InMemoryFileSystem
- InMemoryFileSystem emits `add` event
- ProjectManager checks if the added file is a tsconfig.json => creates new config
- ProjectConfiguration checks if the added file is part of its config => adds it, recompiles

Having this would make other improvements possible, for example doing a streamed workspace/symbol search that fetches & searches files one-by-one, having the ProjectConfiguration add it automatically

PR #233, opened 2017-05-05
Replace createConfigurations() with event API

felixfbecker:
- `InMemoryFileSystem` emits an event whenever a file is added
- Whenever a tsconfig.json is added, ProjectManager creates a new ProjectConfiguration
- The fetching of tsconfig.jsons is already done by `ensureModuleStructure`
- No more `createConfigurations()` call everywhere, no more iterations over all URIs in the workspace

ref #149

### DIFF
--- a/src/project-manager.ts
+++ b/src/project-manager.ts
@@ -1,4 +1,4 @@
-import { Observable } from '@reactivex/rxjs';
+import { Observable, Subscription } from '@reactivex/rxjs';
 import iterate from 'iterare';
 import { Span } from 'opentracing';
 import * as os from 'os';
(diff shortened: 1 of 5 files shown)


src/project-manager.ts
```diff
@@ -97,13 +102,47 @@ export class ProjectManager implements Disposable {
...
+		}
+
+		// Whenever a file with content is added to the InMemoryFileSystem, check if it's a tsconfig.json and add a new ProjectConfiguration
+		this.subscriptions.add(
+			Observable.fromEvent<[string, string]>(inMemoryFileSystem, 'add', Array.of)
+				.filter(([uri, content]) => /\/[tj]sconfig\.json/.test(uri) && !uri.includes('/node_modules/'))
+				.subscribe(([uri, content]) => {
+					const filePath = util.uri2path(uri);
+					let dir = util.toUnixPath(filePath);
+					const pos = dir.lastIndexOf('/');
```
beyang: Can you use `path.posix.dirname`?

felixfbecker: I would like to, but not without changing behaviour here.
```js
> require('path').dirname('')
'.'
> require('path').dirname('/')
'/'
> require('path').dirname('/foo')
'/'
```

felixfbecker:
I had to revert this because it broke xdefinition and I couldn't find out why on first investigation. Any ideas appreciated.

PR #236, opened 2017-05-07
Add prettier symbol kind descriptions to hovers

felixfbecker:
Strips out the `(property)` prefixes (which caused bad syntax highlighting) from the declaration and instead adds them as an extra hover content

image

### DIFF
--- a/src/typescript-service.ts
+++ b/src/typescript-service.ts
@@ -352,13 +352,35 @@ export class TypeScriptService {
 				}
 				const offset: number = ts.getPositionOfLineAndCharacter(sourceFile, params.position.line, params.position.character);
 				const info = configuration.getService().getQuickInfoAtPosition(fileName, offset);
+				this.logger.log(info);
 				if (!info) {
 					return { contents: [] };
 				}
-				const contents: MarkedString[] = [{
-					language: 'typescript',
-					value: ts.displayPartsToString(info.displayParts)
-				}];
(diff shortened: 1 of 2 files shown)


src/typescript-service.ts
```diff
@@ -352,13 +352,35 @@ export class TypeScriptService {
...
-					language: 'typescript',
-					value: ts.displayPartsToString(info.displayParts)
-				}];
+				const contents: (MarkedString | string)[] = [];
+				// Add declaration without the kind
+				const declaration = ts.displayPartsToString(info.displayParts).replace(/^\(.+\)\s+/, '');
+				contents.push({ language: 'typescript', value: declaration });
+				// Add kind with modifiers, e.g. "method (private, ststic)", "class (exported)"
+				if (info.kind) {
+					let kind = '**' + info.kind + '**';
```
beyang: I'd recommend factoring out the computation of `kind` into a separate method for readability's sake.

felixfbecker: I disagree tbh, this is exclusively needed for hover, the method is by no means long and having it here prevents a newcomer having to jump around everywhere to understand what's happening. And it's really just some basic markdown decoration.

beyang: Fair points. In that case, a comment block at the start of the `if` statement would have been helpful to explain what the purpose of the block was.

Issue #218, opened 2017-05-03
Symlinks not followed when mirroring workspace structure in-memory


tomv564:
When adding locally-developed packages as a dependency via `npm link`, the first level of files is found but (eg. `node_modules/mylinkedpackage/index.ts`) but not index.ts's imports further into the tree (like `node_modules/mylinkedpackage/src/myfunctions.ts`).

This results in a broken mylinkedpackage.

Adding `follow: true` to this glob statement allows the resolution to complete successfully:
https://github.com/sourcegraph/javascript-typescript-langserver/blob/master/src/fs.ts#L70

PR #238, opened 2017-05-07
Add follow to glob + test for npm linked packages

tomv564:
For #218, this is to support developers who use npm link to make locally-developed packages available in another locally developed project. 
The existing configuration of 'glob' does not follow symlinks created by `npm link`.

Affects: Users running with LocalFs. 
Risk: `glob` notes the following on the use of `follow` on their documentation:
> Note that this can result in a lot of duplicate references in the presence of cyclic links.

### DIFF
--- a/src/test/fs.test.ts
+++ b/src/test/fs.test.ts
@@ -43,11 +43,7 @@ describe('fs.ts', () => {
 			await fs.writeFile(path.join(projectDir, '@types', 'diff', 'index.d.ts'), 'baz');
 
 			// global package is symolinked into project using npm link
-			if (process.platform === 'win32') {
-				await fs.symlink(somePackageDir, path.join(projectDir, 'node_modules', 'some_package'), 'junction');
-			} else {
-				await fs.symlink(somePackageDir, path.join(projectDir, 'node_modules', 'some_package'));
-			}
+			await fs.symlink(somePackageDir, path.join(projectDir, 'node_modules', 'some_package'), 'junction');
 			fileSystem = new LocalFileSystem(toUnixPath(projectDir));
 		});
 		after(async () => {
--- a/src/test/fs.test.ts
+++ b/src/test/fs.test.ts
@@ -20,15 +20,31 @@ describe('fs.ts', () => {
 			temporaryDir = await new Promise<string>((resolve, reject) => {
 				temp.mkdir('local-fs', (err: Error, dirPath: string) => err ? reject(err) : resolve(dirPath));
 			});
-			baseUri = path2uri('', temporaryDir) + '/';
-			await fs.mkdir(path.join(temporaryDir, 'foo'));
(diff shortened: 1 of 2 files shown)


src/test/fs.test.ts
```diff
@@ -20,15 +20,31 @@ describe('fs.ts', () => {
...
+			await fs.mkdir(path.join(projectDir, '@types'));
+			await fs.mkdir(path.join(projectDir, '@types', 'diff'));
+			await fs.mkdir(path.join(projectDir, 'node_modules'));
+			await fs.writeFile(path.join(projectDir, 'tweedledee'), 'hi');
+			await fs.writeFile(path.join(projectDir, 'tweedledum'), 'bye');
+			await fs.writeFile(path.join(projectDir, 'foo', 'bar.ts'), 'baz');
+			await fs.writeFile(path.join(projectDir, '@types', 'diff', 'index.d.ts'), 'baz');
+
+			// global package is symolinked into project using npm link
+			await fs.symlink(somePackageDir, path.join(projectDir, 'node_modules', 'some_package'));
```
felixfbecker: Afaik, if you don't opt-in in the newest Windows 10 update, Windows requires admin perms to create symlinks. I think this would be better tested in a separate test that is skipped on Windows

tomv564: Thanks for the heads-up, is there an example on ignoring a test on Windows?

felixfbecker: You can do
```ts
(process.platform === 'win32' ? it.skip : it)('should ...', ...)
```

tomv564: I just read that npm creates junctions instead of symlinks on windows (https://github.com/npm/npm/issues/10926), just updated the test.

tomv564:
@felixfbecker tests passed but coverage is 😒

src/test/fs.test.ts
```diff
@@ -20,15 +20,35 @@ describe('fs.ts', () => {
...
+			await fs.mkdir(path.join(projectDir, '@types', 'diff'));
+			await fs.mkdir(path.join(projectDir, 'node_modules'));
+			await fs.writeFile(path.join(projectDir, 'tweedledee'), 'hi');
+			await fs.writeFile(path.join(projectDir, 'tweedledum'), 'bye');
+			await fs.writeFile(path.join(projectDir, 'foo', 'bar.ts'), 'baz');
+			await fs.writeFile(path.join(projectDir, '@types', 'diff', 'index.d.ts'), 'baz');
+
+			// global package is symolinked into project using npm link
+			if (process.platform === 'win32') {
+				await fs.symlink(somePackageDir, path.join(projectDir, 'node_modules', 'some_package'), 'junction');
```
felixfbecker: the third parameter is ignored on other platforms than windows so the if is not needed

tomv564:
@felixfbecker this is also ready for a review, despite the patch coverage being upset.

Issue #220, opened 2017-05-03
File content not pre-loaded for some global declarations 


tomv564:
**tslib**
When using the `importHelpers: true` compiler option and have installed `tslib`, the TS compiler requests content from `node_modules/tslib/tslib.d.ts`. 
This is not recognised as a global declaration path, and it appears that ensureReferencedFiles never receives a reference to it either, so the content is never fetched.
Adding the pattern to `util.ts`'s `globalTSPatterns` made tslib resolve successfully.

**typeRoots and types**
The default typeRoot is at least `node_modules/@types`, we currently limit this to the node directory? (see `/node_modules\/(?:\@|%40)types\/node\/` in `globalTSPatterns`)

Furthermore, adding `typeRoots` to tsconfig's compiler options can override this default.
This means typings from other locations (eg. `typeRoots: ["./mytypes", "../mysharedtypes"]`) should be loaded instead. 

Finally, the `typings` field overrides `typeRoots`, so when specified as `types: ['node']`, then the language server should not need to load content for `node_modules/@types/not_the_node_package/index.d.ts`

See https://www.typescriptlang.org/docs/handbook/tsconfig-json.html

Both these dependencies cannot be properly be discovered until the tsconfig.json is parsed, it might be helpful to move some content-preloading logic to ProjectConfiguration instead.

felixfbecker:
We use a heuristic to determine the global files that are fetched, which means there are always cases that are not fetched. I wonder whether we really have to fetch _all_ files in the workspace instead.

tomv564:
About fetching strategy - could some form of on-demand loading work? (when typescripts asks for a file in the structure for which no content is loaded, just load the content and return that?)
Is the "This file exists but returns no content because it wasn't preloaded" scenario valid or meaningful for both remote and local file systems?

The typeRoots must be used when a developer wants to import shared typings from outside the workspace e.g. from `/company/shared_typings` to `/company/project1`. For now, that developer would be forced to copy the typings or open the whole `/company` folder as workspace.

tomv564:
Outside of typeRoots, I found another example of global typings missing - in this project.
Open any spec in `src/test` and you should see these diagnostics:
```
Cannot find name 'define'.
Cannot find name 'it'.
```
Mocha typings are installed at `node_modules/@types/mocha`, but not referenced in code. As it stands, neither `ensureModuleStructure` nor `ensureReferencedFiles` would load these files.
[...]

PR #240, opened 2017-05-07
Expand isGlobalTSFile + test

tomv564:
Two fixes to allow global typings to load content as documented in #220

As content for these files in not fetched as part of resolving references, the typescript compiler is unable to resolve anything from these typings and will log a diagnostic:
`File '.../node_modules/tslib/tslib.d.ts' is not a module.`

The regex for global typings installed in `node_modules/@typings` was restricted to the 'node' subfolder, this should include all content in @typings.

The TSlib dependency is created in the compiler when the `importHelpers` option is true:
https://github.com/Microsoft/TypeScript/blob/2150a7796b0dfe83c5286755410e0fd1646ed86c/src/compiler/program.ts#L1264-L1273


src/test/util-test.ts
```diff
@@ -35,4 +35,12 @@ describe('util', () => {
 			assert.equal(matches, true);
 		});
 	});
+	describe('isGlobalTSFile()', () => {
+		it('should match declaration files in node_modules', () => {
+			assert.equal(isGlobalTSFile('/node_modules/@types/not_node/not_node.d.ts'), true);
```
felixfbecker: This screws the semantics a bit, `isGlobalTSFile` is supposed to be a heuristic whitelist of known global declarations. Most declarations in `@types` are _not_ global. These should be fetched, but not by changing the definition of `isGlobalTSFile`

tomv564: Ah, right these should match another `import` elsewhere. 
I will remove this case and better document cases where the pre-loading fails (if it still does).

The tsdlib.d.ts is not referenced in source (only ProjectConfiguration can know from checking options.importHelpers) - is it okay if it stays in the isGlobalTSFile check?

felixfbecker: I guess yes since it is global and common

felixfbecker:
Wanna add mocha in here?

tomv564:
Not by itself, no. Isn't whitelisting individual packages a losing race?
I like the idea of fetching all files up-front, then we can perhaps do away with isGlobalTSFile altogether.

felixfbecker:
It is a compromise. We should definitely fetch all files, but that can take a few seconds. We want to be able to respond to a hover or j2d in the first seconds after init without fetching all files. Given most typings are in external module format nowadays, `@types/node` + `@types/mocha` would probably cover 99% of used global typings.

tomv564:
@felixfbecker added mocha, jest and jasmine to isGlobalTS + tests. Want to keep it to this change for now?

PR #243, opened 2017-05-08
Handle Diagnostics without file

### LINKED ISSUE #239: Need to handle diagnostics without file field
The language service is failing to return diagnostics and logging errors when compiler diagnostics are returned alongside file diagnostics.
 
These diagnostics (despite the typing of `ts.Diagnostic`) have no `file` field (typically just a code and a message).

To avoid exceptions handling those diagnostics, adding a  `tsDiagnostics.filter(f => !!f.file)` is needed before calling `convertTsDiagnostic` in `_publishDiagnostics`

felixfbecker:
Closes #239 

@tomv564 does this fix it?

### DIFF
--- a/src/typescript-service.ts
+++ b/src/typescript-service.ts
@@ -896,6 +896,7 @@ export class TypeScriptService {
 		const diagnostics = iterate(tsDiagnostics)
 			// TS can report diagnostics without a file and range in some cases
 			// These cannot be represented as LSP Diagnostics since the range and URI is required
+			// https://github.com/Microsoft/TypeScript/issues/15666
 			.filter(diagnostic => !!diagnostic.file)
 			.map(convertTsDiagnostic)
 			.toArray();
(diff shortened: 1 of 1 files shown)


tomv564:
This looks correct, but I am not able to do the branch gymnastics to test it for you :)
Could consider adding a reference to the GH issue so a future developer may remove the comment if the type is fixed.

felixfbecker:
Okay, will merge in once you give green light this fixes your issue (I don't have a way to repro)

tomv564:
The issue occurs for me as soon as I call didOpen on `typescript-service.ts` in this project:
```
ERROR Handle textDocumentDidOpen: TypeError: Cannot read property 'getLineAndCharacterOfPosition' of undefined
   at convertTsDiagnostic (/Users/tomv/Projects/sourcegraph/javascript-typescript-langserver/lib/diagnostics.js:12:35)
  ...
```
I can confirm your code fixes it! :shipit:

PR #251, opened 2017-05-13
Implement textDocument/rename

felixfbecker:
@tomv564 I split out the rename feature in a separate PR

### DIFF
--- a/src/test/typescript-service-helpers.ts
+++ b/src/test/typescript-service-helpers.ts
@@ -2,14 +2,14 @@ import * as chai from 'chai';
 import * as sinon from 'sinon';
 import * as ts from 'typescript';
 import { CompletionItemKind, CompletionList, TextDocumentIdentifier, TextDocumentItem } from 'vscode-languageserver';
-import { Hover, Location, SignatureHelp, SymbolInformation, SymbolKind } from 'vscode-languageserver-types';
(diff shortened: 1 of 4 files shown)


felixfbecker:
@tomv564 could you give this a review?

tomv564:
Can you explain what function the  JSONPTR serves? I am guessing something to do with streaming unrequested payloads to the client?

felixfbecker:
`JSONPTR` is just a template string tag to help escaping JSON Pointers.

If you mean streaming (which works with JSON Patches) in general, see https://github.com/Microsoft/language-server-protocol/pull/182

felixfbecker:
Does this work for you otherwise?

tomv564:
Can you explain what json pointers mean in the context of the changes being sent to the client? This is the first place that uses them, so I am curious what they point to and why!

 The rest looks OK!

felixfbecker:
Streaming works through JSON Patches, which contain a `path` property that is a JSON Pointer to point to the property in the JSON document to change:

>    Additionally, operation objects MUST have exactly one "path" member.
>    That member's value is a string containing a JSON-Pointer value
>    [RFC6901] that references a location within the target document (the
>    "target location") where the operation is performed.

https://tools.ietf.org/html/rfc6902#section-4

They are already used in every method as the `path` property, but rename is the first method where the path is dynamic because it's a URI -> change dictionary. Since URIs contain slashes, and slashes are reserved in JSON Pointers, they need to be escaped.

tomv564:
Awesome, thanks for educating :)

PR #252, opened 2017-05-13
Implement textDocument/codeAction

felixfbecker:
#251 to be merged first

@tomv564

### DIFF
--- a/src/test/typescript-service-helpers.ts
+++ b/src/test/typescript-service-helpers.ts
@@ -1,8 +1,8 @@
 import * as chai from 'chai';
 import * as sinon from 'sinon';
 import * as ts from 'typescript';
-import { CompletionItemKind, CompletionList, TextDocumentIdentifier, TextDocumentItem } from 'vscode-languageserver';
-import { Hover, Location, SignatureHelp, SymbolInformation, SymbolKind, WorkspaceEdit } from 'vscode-languageserver-types';
+import { CompletionItemKind, CompletionList, DiagnosticSeverity, TextDocumentIdentifier, TextDocumentItem, WorkspaceEdit } from 'vscode-languageserver';
+import { Hover, Location, SignatureHelp, SymbolInformation, SymbolKind } from 'vscode-languageserver-types';
 import { LanguageClient, RemoteLanguageClient } from '../lang-handler';
 import { TextDocumentContentParams, WorkspaceFilesParams } from '../request-type';
 import { SymbolLocationInformation } from '../request-type';
@@ -16,7 +16,7 @@ const assert = chai.assert;
 /**
  * Enforcing strict mode to make tests pass on Windows
  */
-import { setStrict } from '../util';
+import { setStrict, uri2path } from '../util';
 setStrict(true);
 
 export interface TestContext {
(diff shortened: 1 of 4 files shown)


felixfbecker:
Rebased, @tomv564 does this work for you?

src/test/typescript-service-helpers.ts
```diff
@@ -2400,47 +2393,43 @@ export function describeTypeScriptService(createService: TypeScriptServiceFactor
...
-							}
-						}]
+							}]
+						}
 					}
-				}
+				});
 			});
-		} as any);
-	} as any);
```
tomv564: Ah, nice to drop these 👍

src/typescript-service.ts
```diff
@@ -1023,92 +1037,104 @@ export class TypeScriptService {
...
+			}))
+			.startWith({ op: 'add', path: '', value: [] } as AddPatch);
 	}
 
+	/**
+	 * The workspace/executeCommand request is sent from the client to the server to trigger command
+	 * execution on the server. In most cases the server creates a WorkspaceEdit structure and
+	 * applies the changes to the workspace using the request workspace/applyEdit which is sent from
+	 * the server to the client.
+	 */
```
tomv564: Great to see some documentation on how this works!

src/typescript-service.ts
```diff
@@ -991,6 +1038,106 @@ export class TypeScriptService {
...
+						throw new Error(`Expected source file ${change.fileName} to exist in configuration`);
+					}
+					const uri = path2uri(this.root, change.fileName);
+					changes[uri] = change.textChanges.map(({ span, newText }): TextEdit => ({
+						range: {
+							start: ts.getLineAndCharacterOfPosition(sourceFile, span.start),
+							end: ts.getLineAndCharacterOfPosition(sourceFile, span.start + span.length)
+						},
+						newText
+					}));
```
tomv564: So that's where all of `convertTextChange` went. Nice :)

src/typescript-service.ts
```diff
@@ -991,6 +1038,106 @@ export class TypeScriptService {
...
+	 * applies the changes to the workspace using the request workspace/applyEdit which is sent from
+	 * the server to the client.
+	 */
+	workspaceExecuteCommand(params: ExecuteCommandParams, span = new Span()): Observable<OpPatch> {
+		switch (params.command) {
+			case 'codeFix':
+				if (!params.arguments || params.arguments.length < 1) {
+					return Observable.throw(new Error(`Command ${params.command} requires arguments`));
+				}
+				return this.executeCodeFixCommand(params.arguments, span);
```
tomv564: I forgot to highlight this earlier in the PR, but throwing all of `arguments` into `this.executeCodeFixCommand(...` feels a bit dirty. Perhaps the "proper" way is to make sure the FileTextChanges array itself is _in_ `params.arguments[0]`. What do you think?

tomv564:
I just checked out and built this branch and verified both Rename and the codeFix Code Action work

src/typescript-service.ts
felixfbecker: Why do you think it's dirty?

tomv564: What if in the future - besides the array of FileTextChanges - we also need to pass some `Context c` from the CodeAction to `executeCodeFixCommand`.

If the FileTextChanges live in `params.arguments[0]`, then `params.arguments[1]` would be the logical place to put this Context parameter.

If we keep the FileTextChanges in _all of_ `params.arguments`, then where would this extra Context parameter go?

felixfbecker: I would prefer the first parameter to be an object `{ context, fileTextChanges }` then.

tomv564: Yes, that sounds good! 
I guess you are free to modify codeFix's use of Command arguments at any time without compatibility issues, so there is no need to change anything now.

PR #260, opened 2017-05-20
Cross-platform path handling cleanup

tomv564:
Removes util.setStrict and os checks in util
* OS-ness of path is determined purely by checking for drive letters in absolute urls.

Renames `path2uri` to `resolvepath2uri` in hopes of eventually replacing it with a simpler `path2uri`

Adds tests from the whatwg PR

### DIFF
--- a/src/memfs.ts
+++ b/src/memfs.ts
@@ -4,7 +4,7 @@ import * as path from 'path';
 import * as ts from 'typescript';
 import { Logger, NoopLogger } from './logging';
 import { FileSystemEntries, matchFiles } from './match-files';
-import { path2uri, toUnixPath, uri2path } from './util';
+import { resolvepath2uri, toUnixPath, uri2path } from './util';
 
 /**
  * In-memory file cache node which represents either a folder or a file
@@ -129,7 +129,7 @@ export class InMemoryFileSystem extends EventEmitter implements ts.ParseConfigHo
 	 * @param path File path or URI (both absolute or relative file paths are accepted)
 	 */
 	fileExists(path: string): boolean {
-		return this.readFileIfExists(path) !== undefined || this.files.has(path) || this.files.has(path2uri(this.path, path));
+		return this.readFileIfExists(path) !== undefined || this.files.has(path) || this.files.has(resolvepath2uri(this.path, path));
 	}
 
 	/**
@@ -147,7 +147,7 @@ export class InMemoryFileSystem extends EventEmitter implements ts.ParseConfigHo
 	 * If there is no such file, returns undefined
 	 */
 	private readFileIfExists(path: string): string | undefined {
-		const uri = path2uri(this.path, path);
+		const uri = resolvepath2uri(this.path, path);
 		let content = this.overlay.get(uri);
 		if (content !== undefined) {
 			return content;
--- a/src/memfs.ts
+++ b/src/memfs.ts
@@ -4,7 +4,7 @@ import * as path from 'path';
 import * as ts from 'typescript';
 import { Logger, NoopLogger } from './logging';
 import { FileSystemEntries, matchFiles } from './match-files';
-import { resolvepath2uri, toUnixPath, uri2path } from './util';
+import { path2uri, toUnixPath, uri2path } from './util';
 
 /**
  * In-memory file cache node which represents either a folder or a file
@@ -76,6 +76,7 @@ export class InMemoryFileSystem extends EventEmitter implements ts.ParseConfigHo
 			this.files.set(uri, content);
 		}
 		// Add to directory tree
+		// TODO: convert this to use URIs.
 		const filePath = uri2path(uri);
 		const components = filePath.split(/[\/\\]/).filter(c => c);
 		let node = this.rootNode;
@@ -129,7 +130,8 @@ export class InMemoryFileSystem extends EventEmitter implements ts.ParseConfigHo
 	 * @param path File path or URI (both absolute or relative file paths are accepted)
 	 */
 	fileExists(path: string): boolean {
-		return this.readFileIfExists(path) !== undefined || this.files.has(path) || this.files.has(resolvepath2uri(this.path, path));
+		const uri = path2uri(path);
+		return this.files.has(uri) || typeScriptLibraries.has(path);
 	}
 
 	/**
@@ -147,7 +149,7 @@ export class InMemoryFileSystem extends EventEmitter implements ts.ParseConfigHo
 	 * If there is no such file, returns undefined
 	 */
 	private readFileIfExists(path: string): string | undefined {
-		const uri = resolvepath2uri(this.path, path);
+		const uri = path2uri(path);
 		let content = this.overlay.get(uri);
 		if (content !== undefined) {
 			return content;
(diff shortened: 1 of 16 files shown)


tomv564:
Hi Felix, thanks for the review! Agreed on all of your comments, I'll summarise the next steps in the top shortly  as there is quite a bit of cleanup possible from here.

I allowed Windows paths all the way into the in-memory filesystem.
Then I noticed that TS speaks forward slashes (`c:/path/to/my.ts`) back to the host anyways, and converts to forward slashes often as soon as paths are received.
The `ts.resolveModuleName` called in ensureReferences might be an exception, as that broke until I put the `toUnixPath` around the 2nd parameter.

My conclusions are:
* we should probably make sure windows paths are consistently passed as `c:/` to TS
* it might not be worth supporting windows paths in the in-memory filesystem.

felixfbecker:
> I allowed Windows paths all the way into the in-memory filesystem.

The in-memory filesystem works with URIs - could you clarify?

> Then I noticed that TS speaks forward slashes (c:/path/to/my.ts) back to the host anyways, and converts to forward slashes often as soon as paths are received.

I had that suspicion too - very helpful discovery. We should call `toUnixPath()` wherever we interact with TS, but we don't need to prefix with a root slash if I understand correctly. Our `path2uri` functions need to be able to detect Windows paths even if they use forward slashes.

tomv564:
I'm with you on everything, especially making sure we don't mix those
forward-slash paths up with slash-rooted unix paths.

The in-memory file system uses paths to build up the lookup tree IIRC, but
it could become pure URL-based?

felixfbecker:
The InMemoryFileSystem has a directory tree structure to answer TypeScript's `readDirectory` calls. The tree itself of course does not contain slashes as it has one node per directory, and adding to the InMemoryFileSystem should go entirely through URIs now, so that shouldn't be affected and internal to the InMemoryFileSystem.

tomv564:
Gotcha, there is a uri2path in the add method which on windows generates
windows slashes now.
I'll try to eliminate any such conversions in the methods called by
ourselves, and make sure the methods used by TS consistently convert to
path.

Especially fileExists can be improved here, especially if I can prove that
TS only speaks in absolute paths.

src/memfs.ts
```diff
@@ -129,7 +130,8 @@ export class InMemoryFileSystem extends EventEmitter implements ts.ParseConfigHo
 	 * @param path File path or URI (both absolute or relative file paths are accepted)
 	 */
 	fileExists(path: string): boolean {
-		return this.readFileIfExists(path) !== undefined || this.files.has(path) || this.files.has(path2uri(this.path, path));
+		const uri = path2uri(path);
+		return this.files.has(uri) || typeScriptLibraries.has(path);
```
felixfbecker: this doesn't take overlay into account anymore

src/typescript-service.ts
```diff
@@ -1131,7 +1131,7 @@ export class TypeScriptService {
 					if (!sourceFile) {
 						throw new Error(`Expected source file ${change.fileName} to exist in configuration`);
 					}
-					const uri = path2uri(this.root, change.fileName);
+					const uri = path2uri(change.fileName); // resolvepath2uri(this.root, change.fileName);
```
felixfbecker: do you want to keep these comments?

src/memfs.ts
tomv564: Yes, there is another TODO elsewhere about taking the overlay into account. Do you know when that needs to be done?

src/typescript-service.ts
tomv564: Will clean them up!

src/memfs.ts
felixfbecker: yeah, but in this place it was already taken into account because `readFileIfExists` does
Don't know about the other place

src/util.ts
```diff
@@ -62,34 +51,66 @@ export function normalizeUri(uri: string): string {
...
+	pathName = pathName.replace(/\\/g, '/');
+
+	// Windows drive letter must be prefixed with a slash
+	if (pathName[0] !== '/') {
+		pathName = `/${pathName}`;
+	}
+
+	// Escape required characters for path components
+	// See: https://tools.ietf.org/html/rfc3986#section-3.3
+	return encodeURI(`file://${pathName}`).replace(/[?#@]/g, encodeURIComponent);
```
tomv564: I see uriToLocalPath does what you describe here, perhaps I can merge the implementations?

felixfbecker: not sure what you mean by that, this is the path -> URI function, `uriToLocalPath` does URI ->path?

tomv564:
@felixfbecker this branch works for me in both Windows and OS-X ( no thorough testing but I tried a hover, jump to definition, rename in both). Your comments so far should be addressed now, could you have another look?

tomv564:
I'll summarize the assumptions that allowed some of the cleanup here:
* We only pass around absolute paths (and only accept absolute paths from TS)
* TS works with Windows paths, but prefers forward slashes.

Reading the comments above, there is still a small TODO on removing uri2path from memfs's add method.

src/memfs.ts
```diff
@@ -76,8 +76,9 @@ export class InMemoryFileSystem extends EventEmitter implements ts.ParseConfigHo
 			this.files.set(uri, content);
 		}
 		// Add to directory tree
+		// TODO: convert this to use URIs.
```
felixfbecker: we can do this later, I already started an implementation of a URI tree a while ago: https://github.com/sourcegraph/javascript-typescript-langserver/blob/uri-tree/src/trees.ts

src/project-manager.ts
```diff
@@ -672,18 +668,14 @@ export class InMemoryLanguageServiceHost implements ts.LanguageServiceHost {
 	}
 
 	/**
-	 * @param fileName relative or absolute file path
+	 * @param filePath absolute file path
 	 */
-	getScriptSnapshot(fileName: string): ts.IScriptSnapshot | undefined {
-		let exists = this.fs.fileExists(fileName);
-		if (!exists) {
-			fileName = path.posix.join(this.rootPath, fileName);
```
felixfbecker: I wonder whether there was a reason this code was put in place... Are you positive this can only ever be called with an absolute file path?

src/util.ts
```diff
@@ -62,57 +51,42 @@ export function normalizeUri(uri: string): string {
...
+ */
+export function path2uri(path: string): string {
+	// Require a leading slash, on windows prefixed with drive letter
+	if (!/^([a-z]:)?[\\\/]/i.test(path)) {
+		throw new Error(`Expected an absolute uri, got ${path}`);
 	}
-	p = p.split(/[\\\/]/g).map((val, i) => i <= 1 && /^[a-z]:$/i.test(val) ? val : encodeURIComponent(val)).join('/');
-	return normalizeUri(ret + p);
+
+	const [firstSegment, ...remainingSegments] = path.split(/[\\\/]/);
```
felixfbecker: I am changing this since spreading is slow: https://jsperf.com/spread-vs-shift/1

src/project-manager.ts
tomv564: I think this case was reliably absolute paths.
On the other hand, when TS is resolving references a lot more strange paths are requested from the filesystem. I didn't look at TS's logic to understand what it was trying to do, but if we find regressions I would love to document what is happening.

felixfbecker:
Smoke test passed

PR #261, opened 2017-05-22
Add test ensuring project files open without diagnostics

tomv564:
The language service should provide a diagnostic-free experience when opening compiled/linted content from this project.

Diagnostics can indicate type resolution issues which will affect hover/completion/jump-to-definition functionality.
Additionally, this test confirms that all unit-tested parts of the service work correctly together on windows/*nix platforms.
If possible, these tests should not count as coverage towards the project.

Only files from `src/*.ts` are tested for now, but `src/test/*.ts` should be added in the future

### DIFF
--- a/src/test/integration.test.ts
+++ b/src/test/integration.test.ts
@@ -38,7 +38,7 @@ describe.only('running against this project', () => {
 
 			assert.isAtLeast(fileContent.length, 1);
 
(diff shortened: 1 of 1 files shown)


tomv564:
This PR is getting stale, I'll close it until we can make a solid argument for its value.

Issue #208, opened 2017-04-29
Large amount of diagnostics due to module resolution issues


tomv564:
Opening this repo causes a large volume of diagnostics which VS Code / tsc does not show.
```
node_modules/@reactivex/rxjs/typings/globals/node/index.d.ts:
All declarations of 'WeakMap' must have identical type parameters.
```
(caused by Rxjs including typings in the package, fixed by @felixfbecker at https://github.com/ReactiveX/rxjs/issues/2561)

```
src/connection.ts:
File '/Users/tomv/Projects/sourcegraph/javascript-typescript-langserver/node_modules/@reactivex/rxjs/dist/cjs/Rx.d.ts' is not a module.
```
Could this second issue be caused by the TS checker not getting file contents for that path?

felixfbecker:
We could ignore diagnostics from node_modules.

Don't know about the second one.

tomv564:
The `syncProgram` call in ProjectConfiguration's `init` causes the compiler to start requesting modules before `ensureReferencedFiles` has had a chance to load content for them. Commenting it out appears to fix the problem. We could try to remove that sync and make sure other paths force a sync if they don't require references/imports to be loadable.

felixfbecker:
Sounds more like `init` is called to early for me?

tomv564:
Narrowed it down a bit further. Commenting out the sync in `init` helped because it caused `ensureBasicFiles` to early-out until the references were fetched.

Logging reveals that `ensureBasicFiles` is adding the generated declarations (e.g.  lib/lang-handler.d.ts) as source files, which are then picked up by the TS compiler.
Filtering out lib/ in ensureBasic files prevents the diagnostics storm. 

Should `ensureBasicFiles` observe the tsconfig's `includes` directive? Should it even be adding non-global declarations that are not referenced to in some way?

tomv564:
The typescript compiler is also unsuccessfully requesting these two files. This is when the references for a Rxjs have already been resolved.
```
node_modules/@reactivex/rxjs/dist/cjs/add/observable/if.d.ts MISSING 
node_modules/@reactivex/rxjs/dist/cjs/add/observable/throw.d.ts MISSING
```

tomv564:
^^^ This was a false alarm - I was logging empty files as missing.

tomv564:
This is what I see when I didOpen `typescript-service.ts` in this project:

**Compiler diagnostics** (not reported, I added code to log these)
```
[...]

PR #284, opened 2017-06-03
Only add expected declaration files, avoid generated declarations

tomv564:
As discussed in #208, TS will publish diagnostics when it receives empty content when importing a module.

Normally, the content is guaranteed to be in place for file `src/typescript-service.ts` by `ensureReferencedFiles`, but TS was found to be requesting imports like 'vscode-jsonrpc' even when opening an import-free file like 'src/ast.ts'.

The imports were traced back to compiler-emitted declarations like `lib/connection.d.ts` which were exposed to TS by `ensureBasicFiles`.

This PR reduces the adding of declarations to only those matching the project configuration for *source* files.

Besides fixing the repo in #208, this PR allows the tests in #261 to pass and should improve the startup speed of the service on projects with emitted declarations.

### DIFF
--- a/src/project-manager.ts
+++ b/src/project-manager.ts
@@ -11,7 +11,6 @@ import { InMemoryFileSystem } from './memfs';
 import { traceObservable, tracePromise, traceSync } from './tracing';
 import {
 	isConfigFile,
-	isDeclarationFile,
 	isGlobalTSFile,
 	isJSTSFile,
 	isPackageJsonFile,
@@ -745,7 +744,7 @@ export class ProjectConfiguration {
 	 * List of files that project consist of (based on tsconfig includes/excludes and wildcards).
 	 * Each item is a relative file path
 	 */
-	expectedFilePaths: string[];
+	private expectedFilePaths = new Set<string>();
 
 	/**
 	 * @param fs file system to use
@@ -760,7 +759,6 @@ export class ProjectConfiguration {
 		this.versions = versions;
 		this.traceModuleResolution = traceModuleResolution || false;
 		this.rootFilePath = rootFilePath;
-		this.expectedFilePaths = [];
 	}
 
 	/**
@@ -784,7 +782,7 @@ export class ProjectConfiguration {
 		this.service = undefined;
 		this.program = undefined;
 		this.host = undefined;
-		this.expectedFilePaths = [];
+		this.expectedFilePaths = new Set();
 	}
 
 	/**
@@ -847,7 +845,7 @@ export class ProjectConfiguration {
 		}
 		const base = dir || this.fs.path;
 		const configParseResult = ts.parseJsonConfigFileContent(configObject, this.fs, base);
-		this.expectedFilePaths = configParseResult.fileNames;
+		this.expectedFilePaths = new Set(configParseResult.fileNames);
 
 		const options = configParseResult.options;
 		if (/(^|\/)jsconfig\.json$/.test(this.configFilePath)) {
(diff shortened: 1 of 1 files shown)


src/project-manager.ts
```diff
@@ -892,9 +891,14 @@ export class ProjectConfiguration {
...
 
+		// paths have been forward-slashified by TS (see toUnixPath below)
+		// TODO: consider moving expectedFilePaths out of LanguageServiceHost?
+		const expectedFilePaths = this.getHost().expectedFilePaths;
+
 		for (const uri of this.fs.uris()) {
 			const fileName = uri2path(uri);
-			if (isGlobalTSFile(fileName) || (!isDependencyFile(fileName) && isDeclarationFile(fileName))) {
+			if (isGlobalTSFile(fileName) ||
+				(isDeclarationFile(fileName) && expectedFilePaths.includes(toUnixPath(fileName)))) {
```
felixfbecker: This excludes non-declaration files in expected files.

src/project-manager.ts
```diff
@@ -892,9 +891,14 @@ export class ProjectConfiguration {
 			return;
 		}
 
+		// paths have been forward-slashified by TS (see toUnixPath below)
+		// TODO: consider moving expectedFilePaths out of LanguageServiceHost?
```
felixfbecker: Funny, I always thought TS required `expectedFilePath` to be implemented. Turns out it doesn't and it's not even used inside the class. We can totally move it to `ProjectConfiguration`.

src/project-manager.ts
felixfbecker: At first I though this reverted https://github.com/tomv564/javascript-typescript-langserver/commit/021d1b455c8ba48a0f471b96cccb10dd78c23ed5 but it still looks in `this.fs.uris()` 👍 

> - when adding basic files, search for them project-wide instead of searching in expected files list. The reason is that files may be excluded by tsconfig.json rule

src/project-manager.ts
```diff
@@ -749,6 +742,12 @@ export class ProjectConfiguration {
 	private rootFilePath: string;
 
 	/**
+	 * List of files that project consist of (based on tsconfig includes/excludes and wildcards).
+	 * Each item is a relative file path
+	 */
+	expectedFilePaths: string[];
```
felixfbecker: Make it `private` please and initialize here instead of in the constructor

felixfbecker:
Didn't fix https://github.com/sourcegraph/javascript-typescript-langserver/issues/285 unfortunately :(

felixfbecker:
Is it possible to write a unit test for this?

src/project-manager.ts
```diff
@@ -891,18 +889,24 @@ export class ProjectConfiguration {
...
-				(isDeclarationFile(fileName) && this.expectedFilePaths.includes(toUnixPath(fileName)))) {
+			if (isGlobalTSFile(fileName)) {
 				const sourceFile = program.getSourceFile(fileName);
 				if (!sourceFile) {
 					this.getHost().addFile(fileName);
 				}
 			}
 		}
+
+		// Add files that belong to the project according to tsconfig.json settings
```
tomv564: This changes the previous behaviour of only adding declaration files project-wide and adding regular TS files as-needed. Are you sure this is what you want?

felixfbecker: You are correct, will revert

tomv564:
There is likely a bit of set-up effort for a unit test of `ensureBasicFiles`. Might as well cover all the `ensure` functions on ProjectConfiguration! 

The dependencies are also a bit tricky, e.g. `ensureBasicFiles` checks with a live ts.Program instance if a source file is available with adding it to the LanguageServiceHost. Would be a faster and tighter test if all the file-loading interaction could be tested with a mocked LanguageServiceHost. 

I could look at adding these tests in a couple of days, but I'd love to hear what you think of ProjectConfiguration's testability.

felixfbecker:
It would be possible to test the methods in ProjectConfiguration by stubbing all dependencies with sinon and asserting calls, but I don't think there would be much value to it, because it would be very easy to write tests that assert wrong behaviour. The way ensureBasicFiles etc work is a performance trade off (adding the minimal amount of files). I think it makes more sense to have tests that ensure specific hovers/j2d in specific workspaces still work, so we can change the implementation as long as these cases work.

PR #288, opened 2017-06-03
Fix workspace/xreferences

felixfbecker:


src/typescript-service.ts
```diff
@@ -796,35 +803,36 @@ export class TypeScriptService {
...
 												return [symbol];
 											}
 											// If SymbolDescriptor matched and the query contains a PackageDescriptor, get package.json and match PackageDescriptor name
-											// TODO match full PackageDescriptor (version)
+											// TODO match full PackageDescriptor (version) and fill out the symbol.package field
 											const uri = path2uri(definition.fileName);
-											return this._getPackageDescriptor(uri)
-												.mergeMap(packageDescriptor => {
+											return Observable.from(this._getPackageDescriptor(uri, span))
+												.filter(packageDescriptor => !!(packageDescriptor && packageDescriptor.name === params.query.package!.name!))
```
felixfbecker: there should be a test for a query with PackageDescriptor.

PR #323, opened 2017-07-26
Snippet support for autocompletion

tomv564:
If the client sends `snippetSupport: true` in the textDocument/completion/completionItem capability, this PR adds support for responding with these fields set in CompletionItem:

```
insertTextFormat: InsertTextFormat.Snippet,
insertText: 'bar(${1:num})',
```

Both VS Code and Sublime Text support these snippets.
For clients not supporting snippets, CompletionItem will now contain
```
insertTextFormat: InsertTextFormat.PlainText
insertText: 'bar' // same as label
```

### DIFF
--- a/src/typescript-service.ts
+++ b/src/typescript-service.ts
@@ -176,7 +176,6 @@ export class TypeScriptService {
 
 	/**
 	 * Indicates if the client prefers completion results formatted as snippets.
-	 * @type {boolean}
 	 */
 	private supportsCompletionWithSnippets: boolean = false;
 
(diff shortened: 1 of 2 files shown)


src/typescript-service.ts
```diff
@@ -1009,6 +1009,13 @@ export class TypeScriptService {
 					if (entry.kind === 'property') {
 						return entry.name;
 					} else {
+						let index = 0;
+						const parameters = entry.displayParts
+							.filter(p => p.kind === 'parameterName')
+							.map(p => {
+								index++;
```
felixfbecker: `map` passes the index as a second parameter

tomv564:
Great, thanks for the reviews!

Issue #326, opened 2017-08-04
Support Typescript Language Service plugins?


tomv564:
Since TS 2.2.2 there is support for plugins to tsserver.

I would like to be able to run the tslint plugin inside javascript-typescript-langserver, because:
* It re-uses the already-parsed SourceFiles (performance improvement)
* It automatically registers code fixes for lint diagnostics.

As they are implemented as wrappers to ts.LanguageService, all we should need to do is read the relevant settings from tsconfig and apply the plugin decorators whenever we construct a LanguageService.
The whole setup is documented at https://github.com/RyanCavanaugh/sample-ts-plugin.

Do you have any reservations against this? I may put together a PR for this to evaluate the cost

felixfbecker:
Sounds great!

PR #327, opened 2017-08-06
Support TS plugins

tomv564:
An investigation for #326 

There are two ways of specifying plugins to be loaded:

1. `plugins` declaration in **tsconfig.json**
2. `--globalPlugins` argument to tsserver 

There are a few different ways to load the plugin:

1. locally installed with the project (as sibling to typescript itself)
2. global npm package
3. Locations provided via `--pluginProbeLocations`

This PR contains implementation for loading a locally installed plugin specified in **tsconfig.json** only, as documented on Erich Gamma's new vscode plugin at https://github.com/Microsoft/vscode-ts-tslint.

screen shot 2017-08-06 at 20 49 42

Missing:
- [x] Review what parts of PluginCreateInfo need to be wired up for other plugins
- [x] Support for globalPlugins and pluginProbeLocations
- [x] A good way to disable loading plugins altogether for sourcegraph
- [x] Tests


felixfbecker:
Could you add doc comments to all the symbols you added? Makes it easier to review

tomv564:
Plugin support:
The angular and graphql plugins should get all they need from the PluginCreateInfo we provide (languageServer, languageServerHost and project.projectService.logger)

Plugin configuration:
I chose to read settings from initializationOptions as supporting didWorkspaceConfigChange would mean recreating the ProjectConfigurations - I hoped plugin config would be unlikely to change after startup.

Plugin loading:
Plugins installed locally to the project work.
On my machine, plugins are not picked up when installed globally - this may work if javascript-typescript-langserver is installed globally (the peer modules path added to searchPaths).

Security:
If no flags are passed to initializationOptions, no plugin should be loaded unless the user has installed it as a peer package to javascript-typescript-langserver. Is this sufficient?

tomv564:
@felixfbecker the plugins code was extracted and tests have been added, care to have a look?

src/project-manager.ts
```diff
@@ -852,10 +864,39 @@ export class ProjectConfiguration {
...
+		pluginLoader.loadPlugins(options, this.enableProxy.bind(this));
 		this.initialized = true;
 	}
 
 	/**
+	 * Replaces the LanguageService with an instance wrapped by the plugin
+	 * @param pluginModuleFactory function to create the module
+	 * @param configEntry extra settings from tsconfig to pass to the plugin module
+	 */
+	private enableProxy(pluginModuleFactory: PluginModuleFactory, configEntry: ts.PluginImport) {
```
felixfbecker: can this be called multiple times to wrap with multiple plugins?

src/project-manager.ts
```diff
@@ -852,10 +864,39 @@ export class ProjectConfiguration {
 			this.logger
 		);
 		this.service = ts.createLanguageService(this.host, this.documentRegistry);
+		const pluginLoader = new PluginLoader(this.rootFilePath, this.fs, this.initializationOptions, this.logger);
+		pluginLoader.loadPlugins(options, this.enableProxy.bind(this));
```
felixfbecker: please use an arrow function, bind is not type safe

src/project-manager.ts
```diff
@@ -852,10 +864,39 @@ export class ProjectConfiguration {
...
+				project: { projectService: { logger: this.logger }}, // TODO: may need more support
+				languageService: this.getService(),
+				languageServiceHost: this.getHost(),
+				serverHost: {} // TODO: may need an adapter
+			};
+
+			const pluginModule = pluginModuleFactory({ typescript: ts });
+			this.service = pluginModule.create(info);
+		} catch (e) {
+			this.logger.info(`Plugin activation failed: ${e}`);
```
felixfbecker: this should be an error

src/plugins.ts
```diff
@@ -0,0 +1,167 @@
+import * as ts from 'typescript';
+import { Logger, NoopLogger } from './logging';
+import { combinePaths } from './match-files';
+import { InitializationOptions } from './request-type';
+
+// definitions from from TypeScript server/project.ts
```
felixfbecker: could you add a full link to GitHub (with revision)

felixfbecker:
> Plugin configuration:
> I chose to read settings from initializationOptions as supporting didWorkspaceConfigChange would mean recreating the ProjectConfigurations - I believe plugin config would be unlikely to change after startup.

`initializationOptions` is not specced to contain the settings, it is typed as `any`. You should use `didWorkspaceConfigChange`, but you can filter it to only recreate ProjectConfigurations

src/project-manager.ts
tomv564: Yes, every time enableProxy is called, the current `this.service` is wrapped, so the desired result is `plugin2Wrapper(plugin2Wrapper(tsService))`

src/plugins.ts
```diff
@@ -0,0 +1,190 @@
...
+			this.allowLocalPluginLoads = pluginSettings.allowLocalPluginLoads || false;
+			this.globalPlugins = pluginSettings.globalPlugins || [];
+			this.pluginProbeLocations = pluginSettings.pluginProbeLocations || [];
+		}
+	}
+
+	public loadPlugins(options: ts.CompilerOptions, applyProxy: EnableProxyFunc) {
+		// Search our peer node_modules, then any globally-specified probe paths
+		// ../../.. to walk from X/node_modules/javascript-typescript-langserver/lib/project-manager.js to X/node_modules/
+		const searchPaths = [combinePaths(__filename, '../../..'), ...this.pluginProbeLocations];
```
felixfbecker: what is the difference between `combinePaths` and `path.resolve()`?

src/plugins.ts
```diff
@@ -0,0 +1,190 @@
...
+		if (this.globalPlugins) {
+			// Enable global plugins with synthetic configuration entries
+			for (const globalPluginName of this.globalPlugins) {
+				// Skip already-locally-loaded plugins
+				if (!pluginImports || pluginImports.some(p => p.name === globalPluginName)) {
+					continue;
+				}
+
+				// Provide global: true so plugins can detect why they can't find their config
+				this.enablePlugin({ name: globalPluginName, global: true } as ts.PluginImport, searchPaths, applyProxy);
```
felixfbecker: why is `global` not in the typings?

src/plugins.ts
```diff
@@ -0,0 +1,190 @@
...
+	}
+
+	/**
+	 * Tries to load and enable a single plugin
+	 * @param pluginConfigEntry
+	 * @param searchPaths
+	 */
+	private enablePlugin(pluginConfigEntry: ts.PluginImport, searchPaths: string[], enableProxy: EnableProxyFunc) {
+		for (const searchPath of searchPaths) {
+			const resolvedModule =  this.resolveModule(pluginConfigEntry.name, searchPath) as PluginModuleFactory;
```
felixfbecker: why is this cast needed / why is `resolveModule` not typed with that return type?

src/plugins.ts
```diff
@@ -0,0 +1,190 @@
...
+	/**
+	 * Expose resolution logic to allow us to use Node module resolution logic from arbitrary locations.
+	 * No way to do this with `require()`: https://github.com/nodejs/node/issues/5963
+	 * Throws an error if the module can't be resolved.
+	 * stolen from moduleNameResolver.ts because marked as internal
+	 */
+	private resolveJavaScriptModule(moduleName: string, initialDir: string, host: ts.ModuleResolutionHost): string {
+		// TODO: this should set jsOnly=true to the internal resolver, but this parameter is not exposed on a public api.
+		const result =
+			ts.nodeModuleNameResolver(moduleName, /* containingFile */ initialDir.replace('\\', '/') + '/package.json', { moduleResolution: ts.ModuleResolutionKind.NodeJs, allowJs: true }, this.resolutionHost, undefined);
```
felixfbecker: please wrap long lines by making one argument per line:
```ts
const foo = bar(
  baz,
  qux,
  qaz
)
```

src/project-manager.ts
```diff
@@ -852,10 +864,39 @@ export class ProjectConfiguration {
 			this.logger
 		);
 		this.service = ts.createLanguageService(this.host, this.documentRegistry);
+		const pluginLoader = new PluginLoader(this.rootFilePath, this.fs, this.pluginSettings, this.logger);
+		pluginLoader.loadPlugins(options, this.enableProxy.bind(this) /* (factory, config) => this.enableProxy(factory, config)) */);
```
felixfbecker: don't use `bind()`, it's not type safe. wrap in an arrow function instead

src/project-manager.ts
felixfbecker: A name like `wrapService` would make it more clearer to me then, `enable` always sounds like an idempotent function (like setting a boolean to `true`)

src/typescript-service.ts
```diff
@@ -86,9 +87,9 @@ export type TypeScriptServiceFactory = (client: LanguageClient, options?: TypeSc
 /**
  * Settings synced through `didChangeConfiguration`
  */
-export interface Settings {
-	format: ts.FormatCodeSettings;
-}
+export type Settings = {
+	format: ts.FormatCodeSettings
+} & PluginSettings;
```
felixfbecker: prefer `interface Settings extends PluginSettings { ... }`

src/plugins.ts
tomv564: Code is copied from TS project, wanted to keep it as close to original as possible (and we had the function already)

src/plugins.ts
tomv564: Code is copied from TS project - fix should be made there

src/plugins.ts
tomv564: `resolveModule` was not specific to plugin loading, it also was copied and I tried to keep it close to the original.
If it is exposed through the TS api we could remove it from this plugin.ts.

felixfbecker: Is there an open issue for exposing this? If not, could you create one and link it at the top?

src/plugins.ts
felixfbecker: Okay, that wasn't clear to me. Maybe change the wording "based on" at the top of the file to "copied from"?

felixfbecker:
Good to merge once lint failure is fixed

tomv564:
Thanks for the work you put into merging yesterday! I'll try to clear up any of the outstanding issues with Typescript and ping you when the branch has received the possible short-term fixes.

tomv564:
Test is fixed, confirmed it still works on Windows too!
Too bad about the coverage stats, `wrapService` is untested but it's private and only called by successful plugin imports.

Kronuz:
As some plugins also receive configuration options themselves (configFile, ignoreDefinitionFiles, etc., for tslint-language-service), `globalPlugins` could be a list of objects, similarly to the `tsconfig.json` configuration for "compilerOptions" (`globalPlugins` could even be renamed to simply `plugins` to make it the same):

```json
"plugins": [
  {
    "name": "tslint-language-service",
    "alwaysShowRuleFailuresAsWarnings": false,
    "ignoreDefinitionFiles": true,
    "configFile": "../tslint.json",
    "disableNoUnusedVariableRule": false,
    "supressWhileTypeErrorsPresent": false,
    "mockTypeScriptVersion": false
  }
]
```

I'd suggest a way to be able to pass those to the plugin maybe passing the whole object here, prior maybe adding `global: true` to it:

https://github.com/sourcegraph/javascript-typescript-langserver/blob/54304b814a0602457a9c66afea7dc20753b7c509/src/plugins.ts#L122

What do you think?

PR #343, opened 2017-08-29
Add typeroots support

tomv564:
On codebases that rely on type roots for distributing extra declaration files, this code should prevent diagnostics about missing types and improve completions/hovers/etc.

To do: 
- [x] Fix lint error
- [x] Add test(s)

### DIFF
--- a/src/project-manager.ts
+++ b/src/project-manager.ts
@@ -735,6 +735,11 @@ export class ProjectConfiguration {
 	private expectedFilePaths = new Set<string>();
 
 	/**
+	 * List of resolved extra root directories to allow global typings to be loaded from.
+	 */
+	private typeRoots: string[];
+
+	/**
 	 * @param fs file system to use
 	 * @param documentRegistry Shared DocumentRegistry that manages SourceFile objects
 	 * @param rootFilePath root file path, absolute
@@ -838,6 +843,10 @@ export class ProjectConfiguration {
 		this.expectedFilePaths = new Set(configParseResult.fileNames);
 
 		const options = configParseResult.options;
+		this.typeRoots = options.typeRoots ?
+			options.typeRoots.map((r: string) => path.resolve(this.rootFilePath, r)) :
+			[];
+
 		if (/(^|\/)jsconfig\.json$/.test(this.configFilePath)) {
 			options.allowJs = true;
 		}
@@ -864,6 +873,17 @@ export class ProjectConfiguration {
 
 	private ensuredBasicFiles = false;
 
+	private isTypeRootDeclaration(path: string) {
+		if (isDeclarationFile(path)) {
+			for (const base of this.typeRoots) {
+				if (path.startsWith(base)) {
+					return true;
+				}
+			}
+		}
+		return false;
+	}
+
 	/**
 	 * Ensures we added basic files (global TS files, dependencies, declarations)
 	 */
@@ -882,7 +902,9 @@ export class ProjectConfiguration {
 		// Add all global declaration files from the workspace and all declarations from the project
 		for (const uri of this.fs.uris()) {
 			const fileName = uri2path(uri);
-			if (isGlobalTSFile(fileName) || (isDeclarationFile(fileName) && this.expectedFilePaths.has(toUnixPath(fileName)))) {
+			if (isGlobalTSFile(fileName) ||
+				this.isTypeRootDeclaration(fileName) ||
+				(isDeclarationFile(fileName) && this.expectedFilePaths.has(toUnixPath(fileName)))) {
 				const sourceFile = program.getSourceFile(fileName);
 				if (!sourceFile) {
 					this.getHost().addFile(fileName);
--- a/src/project-manager.ts
+++ b/src/project-manager.ts
@@ -880,7 +880,7 @@ export class ProjectConfiguration {
 	private isExpectedDeclarationFile(fileName: string) {
 		if (isDeclarationFile(fileName)) {
 			return this.expectedFilePaths.has(toUnixPath(fileName)) ||
-					this.typeRoots.some(root => fileName.startsWith(root))
+					this.typeRoots.some(root => fileName.startsWith(root));
 		} else {
 			return false;
 		}
--- a/src/project-manager.ts
+++ b/src/project-manager.ts
@@ -338,6 +338,10 @@ export class ProjectManager implements Disposable {
 		});
 	}
 
+	/**
+	 * Determines if a tsconfig/jsconfig needs additional declaration files loaded.
+	 * @param filePath
+	 */
 	isConfigDependency(filePath: string): boolean {
 		for (const config of this.configurations()) {
 			config.ensureConfigFile();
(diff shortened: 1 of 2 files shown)


tomv564:
@felixfbecker, I am a bit stuck on this test - the types.d.ts file is added to the in-memory file system but its content is not loaded. What am I missing?

tomv564:
@felixfbecker The content-preload for declaration files was missing typeRoots, tests pass with windows and unix-style paths now.
I'm not happy with having introduced yet another `ensure` function, but couldn't find any other place that loads content based on tsconfig settings?

src/project-manager.ts
```diff
@@ -337,6 +338,24 @@ export class ProjectManager implements Disposable {
...
+		}
+		return false;
+	}
+
+	ensureConfigDependencies(): Observable<never> {
+		return observableFromIterable(this.inMemoryFs.uris())
+			.filter(uri => this.isConfigDependency(uri2path(uri)))
+			.mergeMap(uri => this.updater.ensure(uri))
+			.publishReplay()
+			.refCount();
```
felixfbecker: why publishReplay + refCount?

src/project-manager.ts
```diff
@@ -838,6 +862,11 @@ export class ProjectConfiguration {
 		this.expectedFilePaths = new Set(configParseResult.fileNames);
 
 		const options = configParseResult.options;
+		const pathResolver = /[a-z]:\//i.test(base) ? path.win32 : path.posix;
```
felixfbecker: the regexp should only match at the beginning of the string

src/project-manager.ts
```diff
@@ -865,6 +894,19 @@ export class ProjectConfiguration {
 	private ensuredBasicFiles = false;
 
 	/**
+	 * [isExpectedDeclarationFile description]
+	 * @param {string} fileName [description]
```
felixfbecker: template?

src/project-manager.ts
tomv564: I looked at ensureModuleStructure, which I also understood would be invoked many times but should always return the same (cached) sequence. But I'd love to know how this use case differs (and remove it if unnecessary).

tomv564:
@felixfbecker: would you have time to look at this PR again soon?

src/project-manager.ts
felixfbecker: this doesn't actually save the (i.e. cache) the Observable, so it doesn't has a use here. If you want to always return the same Observable you need to save it in a property

src/project-manager.ts
```diff
@@ -735,6 +754,11 @@ export class ProjectConfiguration {
 	private expectedFilePaths = new Set<string>();
 
 	/**
+	 * List of resolved extra root directories to allow global type declaration files to be loaded from.
```
felixfbecker: Is this restricted to global typings?

src/project-manager.ts
```diff
@@ -735,6 +754,11 @@ export class ProjectConfiguration {
...
+	 */
+	private typeRoots: string[];
```
felixfbecker: is the order important? if not, use a Set

src/project-manager.ts
```diff
@@ -865,6 +894,19 @@ export class ProjectConfiguration {
...
+	 * Determines if a fileName is a declaration file within expected files or type roots
+	 * @param fileName
+	 */
+	public isExpectedDeclarationFile(fileName: string) {
+		if (isDeclarationFile(fileName)) {
+			return this.expectedFilePaths.has(toUnixPath(fileName)) ||
+					this.typeRoots.some(root => fileName.startsWith(root));
+		} else {
+			return false;
+		}
```
felixfbecker: this function just needs to be a single boolean `return`

src/project-manager.ts
```diff
@@ -865,6 +894,19 @@ export class ProjectConfiguration {
 	private ensuredBasicFiles = false;
 
 	/**
+	 * Determines if a fileName is a declaration file within expected files or type roots
+	 * @param fileName
+	 */
+	public isExpectedDeclarationFile(fileName: string) {
```
felixfbecker: return type

src/project-manager.ts
```diff
@@ -337,6 +338,24 @@ export class ProjectManager implements Disposable {
 		});
 	}
 
+	isConfigDependency(filePath: string): boolean {
```
felixfbecker: please add a docblock

src/project-manager.ts
```diff
@@ -337,6 +338,24 @@ export class ProjectManager implements Disposable {
...
+		for (const config of this.configurations()) {
+			config.ensureConfigFile();
+			if (config.isExpectedDeclarationFile(filePath)) {
+				return true;
+			}
+		}
+		return false;
+	}
+
+	ensureConfigDependencies(): Observable<never> {
```
felixfbecker: please add a docblock

src/project-manager.ts
tomv564: From https://www.typescriptlang.org/docs/handbook/tsconfig-json.html:
> By default all visible “@types” packages are included in your compilation.
> If typesRoots is specified, only packages under typeRoots will be included.
> Keep in mind that automatic inclusion is only important if you’re using files with global declarations (as opposed to files declared as modules)

src/project-manager.ts
tomv564: Not sure if `this.typeRoots.some(root => fileName.startsWith(root));` is a good fit for Set?

src/project-manager.ts
tomv564: They clarify in that page as well: 
> If you use an import "foo" statement, for instance, TypeScript may still look through node_modules & node_modules/@types folders to find the foo package.

tomv564:
@felixfbecker: your suggestions have been applied (with the exception of Set, see comment)

tomv564:
Ping @felixfbecker, congrats on the big release today. Any chance we can work towards merging this PR and the plugin support PR?

felixfbecker:
Thank you, as you can imagine I was very busy in the last days :) Taking a look

src/project-manager.ts
```diff
@@ -321,6 +321,7 @@ export class ProjectManager implements Disposable {
 			span.addTags({ uri, maxDepth });
 			ignore.add(uri);
 			return this.ensureModuleStructure(span)
+				.concat(Observable.defer(() => this.ensureConfigDependencies()))
```
felixfbecker: Does this have a performance impact? I think it probably has, as you are iterating in `ensureConfigDependencies()` over every single file in the workspace, and `ensureReferencedFiles()` is called recursively for every referenced file, so you are repeating that iteration every time. I think `ensureConfigDependencies()` needs to memoize its return value in the same way `ensureAllFiles()` does, or you could just add that logic to `ensureModuleStructure()` (which would already is memoized and also has memoize cache invalidation logic).

tomv564:
@felixfbecker I applied your suggestion, and verified that:
* The logic still works with a `typeRoots` dependent project
* Hovers on the Microsoft/Typescript source are still appearing quickly

PR #344, opened 2017-09-04
Switch to resolving expensive completionItem details

tomv564:
Requesting extra completion details was found to be very expensive in #331, this PR will require clients to request these details via a separate request.

To do:
- [x] Clean up commented code

### DIFF
--- a/src/test/typescript-service-helpers.ts
+++ b/src/test/typescript-service-helpers.ts
@@ -5,7 +5,7 @@ import { applyReducer, Operation } from 'fast-json-patch';
 import { IBeforeAndAfterContext, ISuiteCallbackContext, ITestCallbackContext } from 'mocha';
 import * as sinon from 'sinon';
 import * as ts from 'typescript';
-import { CompletionItemKind, CompletionList, DiagnosticSeverity, InsertTextFormat, TextDocumentIdentifier, TextDocumentItem, WorkspaceEdit } from 'vscode-languageserver';
+import { CompletionItemKind, CompletionList, DiagnosticSeverity, InsertTextFormat, TextDocumentIdentifier, TextDocumentItem, WorkspaceEdit, CompletionItem } from 'vscode-languageserver';
 import { Command, Diagnostic, Hover, Location, SignatureHelp, SymbolInformation, SymbolKind } from 'vscode-languageserver-types';
 import { LanguageClient, RemoteLanguageClient } from '../lang-handler';
 import { DependencyReference, PackageInformation, ReferenceInformation, TextDocumentContentParams, WorkspaceFilesParams } from '../request-type';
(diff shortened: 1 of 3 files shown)


src/test/typescript-service-helpers.ts
```diff
@@ -2169,44 +2169,22 @@ export function describeTypeScriptService(createService: TypeScriptServiceFactor
...
-				{
-					label: 'qux',
-					kind: CompletionItemKind.Property,
-					documentation: 'qux doc',
-					sortText: '0',
-					insertTextFormat: InsertTextFormat.Snippet,
-					insertText: 'qux',
-					detail: '(property) A.qux: number'
-				}
-			]);
```
felixfbecker: why is this assertion removed?

src/test/typescript-service-helpers.ts
```diff
@@ -2169,44 +2169,22 @@ export function describeTypeScriptService(createService: TypeScriptServiceFactor
...
+
+			assert.deepEqual(resolved, {
+				label: 'bar',
+				kind: CompletionItemKind.Method,
+				documentation: 'bar doc',
+				insertText: 'bar(${1:num})',
+				insertTextFormat: InsertTextFormat.Snippet,
+				sortText: '0',
+				detail: '(method) A.bar(num: number): number'
+			});
```
felixfbecker: this would better be a second test

src/typescript-service.ts
```diff
@@ -1022,26 +1022,14 @@ export class TypeScriptService {
...
-								item.insertText = details.name;
-							}
-						}
+
+						// context for future resolve requests:
+						item.data = {
+							uri,
+							offset,
+							entryName: entry.name
+						};
```
felixfbecker: could you define an interface for this / add this to a custom CompletionItem interface in request-type?

src/typescript-service.ts
```diff
@@ -1050,6 +1038,49 @@ export class TypeScriptService {
...
+							item.insertText = details.name + `(${paramString})`;
+						} else {
+							item.insertText = details.name;
+
+						}
+					} else {
+						item.insertTextFormat = InsertTextFormat.PlainText;
+						item.insertText = details.name;
+					}
+					delete item.data;
```
felixfbecker: `delete` is slow, is this needed?

src/test/typescript-service-helpers.ts
tomv564: The snippets logic is only used when resolving completion items, and I thought it would be a lot of logic to resolve all the items from the initial response.
Now that I think of it, it's probably not that bad - I'll give it a try!

felixfbecker: I would leave this test as is (just change the objects) and add a separate test for the resolving

src/typescript-service.ts
tomv564: It's not needed. It would be nice to reduce payload and not have to include the field in the asserts on resolved completion items (I wonder if the tests should even care at all what is in the data field?). Could we set to `undefined`, `null`, or should I just leave it be?

tomv564:
I didn't split the snippets test cases yet, and assigned undefined to the data field after resolving for now. Thanks for the comments!

tomv564:
@felixfbecker: your review comments are addressed, commented code is removed.

One snippet improvement was also included:
Previously all completions that were not `property` returned `function(arguments..)` style snippets, which is clearly not appropriate for e.g. `Interface` completions. The function-style snippets are now only returned for `method` and `function` completion item kinds.

PR #345, opened 2017-09-04
Delay diagnostics to give priority to other client requests

tomv564:
This should remove an initial delay before the language server answers a completion request (see #331)
Currently uses setTimeout, would like to know about a preferable approach?

### DIFF
--- a/src/typescript-service.ts
+++ b/src/typescript-service.ts
@@ -1267,7 +1267,7 @@ export class TypeScriptService {
 		// Ensure files needed for most operations are fetched
 		await this.projectManager.ensureReferencedFiles(uri).toPromise();
 		this.projectManager.didOpen(uri, params.textDocument.text);
-		this._publishDiagnostics(uri);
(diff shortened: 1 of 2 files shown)


src/typescript-service.ts
```diff
@@ -1267,7 +1267,7 @@ export class TypeScriptService {
 		// Ensure files needed for most operations are fetched
 		await this.projectManager.ensureReferencedFiles(uri).toPromise();
 		this.projectManager.didOpen(uri, params.textDocument.text);
-		this._publishDiagnostics(uri);
+		setTimeout(() => this._publishDiagnostics(uri), 200);
```
felixfbecker: if `publishDiagnostics` throws this crashes the process. I would do this:
```ts
await new Promise(resolve => setTimeout(resolve, 200))
this._publishDiagnostics(uri)
```

tomv564: Whoops, all error handling out the window.
Cleaned up the tests after applying your suggestion, thanks for the help!

PR #356, opened 2017-09-21
Upgrade to typescript 2.4.2

tomv564:
Upgrading directly to 2.5 yielded even more issues, this smaller bump should be more manageable.

Compile errors remaining:
- [x] Removed types/fields in ast.ts 
- [x] Element type in Observables is now strictly checked causing issues with `Observable<never>`s to which we concat  `Observable<DesiredResultType>`

Test issues remaining:
- [x] Fix tests around imports
- [x] Fix test and runtime issues around imports/references


felixfbecker:
> Removed types/fields in ast.ts

I don't know tbh why ast.ts doesn't simply use `getChildren()`.

tomv564:
@felixfbecker: Only the typing issue with Observables remains, might be an easy fix for you? I got lost in RxJS operators last night, perhaps replacing `concat` with `toArray().mergeMap()` works?

Some notes for future review:
* Needed to `ensureSourceFile` on resolved references as TS's symbol lookups were not being populated by `ensureReferencedFiles` anymore. Would be good peace-of-mind to know what changed.
* Tests pass with `getChildren()` in ast.ts, is the test coverage sufficient to move on?

tomv564:
An update on the imports issue: 
I added `ensureSourceFile` to `ensureReferencedFiles` earlier as TS was no longer parsing references automatically. This introduced a timing issue where TS starts asking for other imports _before_ ensureReferencedFiles has reached them.
I have two ideas left to pursue:
* Investigate what changed in TS so it doesn't parse referenced files automatically as it did pre-2.4
* Run `addSourceFile` on all the referenced files after all the whole tree is resolved.

tomv564:
ping @felixfbecker: I think we are ready for a review on this.

Today's update:
* Found out `import` resolution was failing because we didn't implement the _optional_ `fileExists` / `readFile` on LanguageServerHost -> created an issue about this (linked above).
* Removed the `ensureSourceFile` hack and other unneeded test fixes.

tomv564:
Upgrading to TS 2.5.2 showed new compile errors (packages.ts also needs the `pairOf` fix), but a few completion tests started failing, too. Not a freebie, sadly.

felixfbecker:
Awesome, I'll take a look when I find time

PR #362, opened 2017-10-03
Fix global type loading on Windows paths

tomv564:
This PR now contains 3 fixes for Windows path handling caused by two inconsistencies

The first inconsistency existed between `expectedFiles` (unix paths) and `typeRoots` (windows paths)

**Typings from e.g. jasmine/mocha not resolved by Typescript**

Repro: Open any test file in this project, receive diagnostics about not finding `it` and `describe`
Cause: Files were loaded in memory, but `ensureBasicFiles` passed windows paths into unix-style path filters (`isGlobalTsFile`).
Fix: Convert to unix paths before passing to filter methods in `ensureBasicFiles` (https://github.com/sourcegraph/javascript-typescript-langserver/commit/7f57e21bc71fa02b9533ebe0fafeb608a5ab7dd9)
Test: ✅ Added in https://github.com/sourcegraph/javascript-typescript-langserver/pull/362/commits/6d9d3d0f34d5354e924d25468cbd2a15376fec30

After this fix, `isExpectedDeclarationFile` now expected a unixPath, causing a new issue:

**Typings from typeRoots not resolved by Typescript**

Repro: Open a project using typeRoots, receive diagnostics about not finding types from typeRoots
Cause: Files from typeRoots were not loaded into memory by `ensureConfigDependencies` as it was still passing Windows paths into the now unix-path-based `isConfigDependency`
Fix: Make `typeRoots` unix-path based so it is consistent with checks against `expectedFiles`. (https://github.com/sourcegraph/javascript-typescript-langserver/commit/f0c8e02fdd58465e415adbd673151a45952bdf38)
Test: ✅ Updated to check host in https://github.com/sourcegraph/javascript-typescript-langserver/pull/362/commits/6d9d3d0f34d5354e924d25468cbd2a15376fec30

This fixed the type resolution issues I had been seeing while working on Windows. :tada:

To add test coverage, I made the ProjectManager tests use different rootUris. This did not cover the above issues, but it did uncover another inconsistency:

**Windows `rootPath` not handled correctly for configuration lookups in ProjectManager**

When ProjectManager is constructed with a Windows `rootPath`:
* It is not trimmed correctly (see `trimmedRootPath`)
* It puts Windows paths into `this.configs` which otherwise uses unix paths. 

Impact: My guess is default configurations did not work on windows, as they were windows-style and the rest of the code assumed unix-style paths were used..
Fix: Handle `rootPath` and configuration lookups in a platform-sensitive way (because there is no need to normalise to unix paths here?)

### DIFF
--- a/src/project-manager.ts
+++ b/src/project-manager.ts
@@ -24,6 +24,8 @@ import {
     uri2path
 } from './util'
 
+const LastForwardOrBackwardSlash = /[\\\/][^\\\/]*$/
+
 /**
  * Implementaton of LanguageServiceHost that works with in-memory file system.
  * It takes file content from local cache and provides it to TS compiler on demand
@@ -622,7 +624,7 @@ export class ProjectManager implements Disposable {
                 .filter(([uri, content]) => !!content && /\/[tj]sconfig\.json/.test(uri) && !uri.includes('/node_modules/'))
                 .subscribe(([uri, content]) => {
                     const filePath = uri2path(uri)
-                    const pos = filePath.includes('\\') ? filePath.lastIndexOf('\\') : filePath.lastIndexOf('/')
+                    const pos = filePath.search(LastForwardOrBackwardSlash)
                     const dir = pos <= 0 ? '' : filePath.substring(0, pos)
                     const configType = this.getConfigurationType(filePath)
                     const configs = this.configs[configType]
@@ -937,7 +939,7 @@ export class ProjectManager implements Disposable {
             if (config) {
                 return config
             }
-            const pos = dir.includes('\\') ? dir.lastIndexOf('\\') : dir.lastIndexOf('/')
+            const pos = dir.search(LastForwardOrBackwardSlash)
             if (pos <= 0) {
                 dir = ''
             } else {
--- a/src/project-manager.ts
+++ b/src/project-manager.ts
@@ -589,7 +589,7 @@ export class ProjectManager implements Disposable {
 
         // Create catch-all fallback configs in case there are no tsconfig.json files
         // They are removed once at least one tsconfig.json is found
-        const trimmedRootPath = this.rootPath.replace(/\/+$/, '')
+        const trimmedRootPath = this.rootPath.replace(/[\\\/]+$/, '')
         const fallbackConfigs: {js?: ProjectConfiguration, ts?: ProjectConfiguration} = {}
         for (const configType of ['js', 'ts'] as ConfigType[]) {
             const configs = this.configs[configType]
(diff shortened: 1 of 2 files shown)


tomv564:
@felixfbecker would you want to have a look at this? Not a clean diff, but I hope the comments above help explain.

The main design issue I tried to solve is knowing when windows paths are converted to unix paths, my guidelines for solving the inconsistency were:
* ProjectManager uses Windows paths as it interacts more with the filesystem.
* ProjectConfiguration uses unixPaths as it works closely with the typescript configuration/library which use unix paths on windows.

src/project-manager.ts
```diff
@@ -929,19 +925,19 @@ export class ProjectManager implements Disposable {
...
         }
-        const rootPath = this.rootPath.replace(/\/+$/, '')
+        const rootPath = this.rootPath.replace(/[\\\/]+$/, '')
         while (dir && dir !== rootPath) {
             config = configs.get(dir)
             if (config) {
                 return config
             }
-            const pos = dir.lastIndexOf('/')
+            const pos = dir.includes('\\') ? dir.lastIndexOf('\\') : dir.lastIndexOf('/')
```
felixfbecker: I'd use `string.search()`

src/test/project-manager.test.ts
```diff
@@ -3,134 +3,177 @@ import chaiAsPromised = require('chai-as-promised')
...
 import { InMemoryFileSystem } from '../memfs'
 import { ProjectManager } from '../project-manager'
+import { uri2path } from '../util'
 import { MapFileSystem } from './fs-helpers'
 chai.use(chaiAsPromised)
 const assert = chai.assert
 
 describe('ProjectManager', () => {
+    for (const rootUri of ['file:///', 'file:///c:/foo/bar/', 'file:///foo/bar/']) {
+        testWithRootUri(rootUri)
```
felixfbecker: why not inline this here?

felixfbecker:
> * ProjectManager uses Windows paths as it interacts more with the filesystem.
> * ProjectConfiguration uses unixPaths as it works closely with the typescript configuration/library which use unix paths on windows.

Could you make this documentation available in the source code somewhere, e.g. in the docblock of the respective classes?

src/project-manager.ts
tomv564: Definitely more compact:
```javascript
filePath.search(/[\\\/][^\\\/]*$/)
```
Do you prefer this? If so, I propose storing the regex as a static member somewhere (or adding this functionality to `util.ts`.

tomv564:
For visibility I propose adding it on every docblock that describes a field/parameter.

I wish typing could help with this, eg
```typescript
type UnixPath: string
```
An option is to use `interface UnixPath extends string`, but this requires code changes.
What do you prefer?

src/test/project-manager.test.ts
tomv564: Fixed (if I understood correctly)

tomv564:
@felixfbecker your suggestions were applied, ready for your review again!

src/project-manager.ts
```diff
@@ -24,6 +24,8 @@ import {
     uri2path
 } from './util'
 
+const LastForwardOrBackwardSlash = /[\\\/][^\\\/]*$/
```
felixfbecker: use CONSTANT_CASE

Issue #370, opened 2017-10-14
workspace/executeCommand on code fixes fails on Windows


tomv564:
Unexpected unix-style path from typescript being fed into windows-path expecting code:

```{'data': {}, 'code': -32001, 'message': 'TypeScript config file for C:/Users/tvanommeren/someproject/somefile.ts not found'}```

On a request:

```
{ 
  title: 'Add ListItem to existing import declaration from.',
  arguments: [ { fileName: 'C:/Users/tvanommeren/someproject/somefile.ts, textChanges: [ { span: { length: 0, start: 252 }, newText: ', ListItem' } ] } ],
  command: 'codeFix' 
}
```

PR #377, opened 2017-10-22
Fix preloading of files from code fixes on Windows

tomv564:
For #370, where getConfiguration now takes windows paths.
The paths from typescript need to be converted from forward slashes back to windows paths.
Small fix, do you want a test or discuss normalised paths?


src/typescript-service.ts
```diff
@@ -1245,7 +1245,9 @@ export class TypeScriptService {
 
         return this.projectManager.ensureOwnFiles(span)
             .concat(Observable.defer(() => {
-                const configuration = this.projectManager.getConfiguration(fileTextChanges[0].fileName)
+                // Configuration lookup uses Windows paths, FileTextChanges uses unix paths. Convert to backslashes.
+                const firstChangedFile = uri2path(path2uri(fileTextChanges[0].fileName))
```
felixfbecker: If the purpose of this is to convert to backslashes, why not simply do `filename.replace(/\//g, '\\')`? Converting to a URI and back seems like a hack

felixfbecker: I see that the intention is to only do it if it's a path with a drive letter, I'd still prefer to leave the URI functions out of this

tomv564: Yes it's not the cleanest looking solution :) I'll extract the needed logic from uri2path.

tomv564: Done.

PR #384, opened 2017-10-26
Empty handler for initialized notification

tomv564:
The language server currently logs a warning when the `initialized` notification is received.

As we are not dynamically registering capabilities on the client, there should be nothing to do but silently handle the notification?

See: https://github.com/Microsoft/language-server-protocol/blob/master/protocol.md#new-initialized-notification

### DIFF
--- a/src/typescript-service.ts
+++ b/src/typescript-service.ts
@@ -1324,6 +1324,10 @@ export class TypeScriptService {
             .startWith({ op: 'add', path: '', value: { changes: {} } as WorkspaceEdit } as Operation)
     }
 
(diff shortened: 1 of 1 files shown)


src/typescript-service.ts
```diff
@@ -1324,6 +1324,10 @@ export class TypeScriptService {
             .startWith({ op: 'add', path: '', value: { changes: {} } as WorkspaceEdit } as Operation)
     }
 
+    public async initialized(): Promise<void> {
```
felixfbecker: please copy over the description

PR #415, opened 2017-12-05
fix(codefix): correct hover information for type signatures with callbacks

### LINKED ISSUE #413: Hover information for type signatures with callbacks is incorrect
https://github.com/sourcegraph/javascript-typescript-langserver/blob/d02f2340b22000cfdcdace7c32d1dad71a439063/src/typescript-service.ts#L612

This should be a non-greedy `+`, as in `\(.+?\)`.
Consider the string
```
(local function) act(handle: () => void): void
``` 
Right now this is incorrectly chopped to 
```
 => void): void
```
instead of the correct
```
act(handle: () => void): void
``` 
HTH

danr:
Fixes #413


felixfbecker:
Could you add a test for this to prevent regressions?

danr:
I cannot read the output from `tslint`

felixfbecker:
Just run `npm run prettier`

danr:
Wohoo! :tada:

PR #430, opened 2018-02-15
feat(typescript): upgrade to typescript 2.6.2

daviwil:
This change updates TypeScript to 2.6.2, the latest stable version.  We had to make some changes to address [API breaking changes](https://github.com/Microsoft/TypeScript/wiki/API-Breaking-Changes#typescript-26) and a compiler change which required the code [here](https://github.com/sourcegraph/javascript-typescript-langserver/pull/430/files#diff-7a192e75048b31deb4cc6b82f30f91e8R105) to change to avoid the use of `constructor` as an object key.

Please let us know if there's anything else we can change to get this accepted!

Resolves https://github.com/sourcegraph/javascript-typescript-langserver/pull/359

/cc @damieng

### DIFF
--- a/package.json
+++ b/package.json
@@ -51,7 +51,7 @@
     "rxjs": "^5.5.0",
     "semaphore-async-await": "^1.5.1",
     "string-similarity": "^1.1.0",
-    "typescript": "2.4.2",
+    "typescript": "2.6.2",
     "vscode-jsonrpc": "^3.3.1",
     "vscode-languageserver": "^3.1.0",
     "vscode-languageserver-types": "^3.0.3"
--- a/src/typescript-service.ts
+++ b/src/typescript-service.ts
@@ -100,25 +100,25 @@ export interface Settings extends PluginSettings {
 /**
  * Maps string-based CompletionEntry::kind to enum-based CompletionItemKind
  */
-const completionKinds: { [name: string]: CompletionItemKind } = {
-    class: CompletionItemKind.Class,
-    constructor: CompletionItemKind.Constructor,
-    enum: CompletionItemKind.Enum,
-    field: CompletionItemKind.Field,
-    file: CompletionItemKind.File,
-    function: CompletionItemKind.Function,
-    interface: CompletionItemKind.Interface,
-    keyword: CompletionItemKind.Keyword,
-    method: CompletionItemKind.Method,
-    module: CompletionItemKind.Module,
-    property: CompletionItemKind.Property,
-    reference: CompletionItemKind.Reference,
-    snippet: CompletionItemKind.Snippet,
-    text: CompletionItemKind.Text,
-    unit: CompletionItemKind.Unit,
-    value: CompletionItemKind.Value,
-    variable: CompletionItemKind.Variable,
-}
+const completionKinds = new Map<string, CompletionItemKind>([
+    [`class`, CompletionItemKind.Class],
+    [`constructor`, CompletionItemKind.Constructor],
+    [`enum`, CompletionItemKind.Enum],
+    [`field`, CompletionItemKind.Field],
+    [`file`, CompletionItemKind.File],
+    [`function`, CompletionItemKind.Function],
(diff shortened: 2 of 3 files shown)


daviwil:
Looks like there are some tests which depend on code locations in a specific version of TypeScript, will get those fixed momentarily.

felixfbecker:
Yes, some tests assert on go-to-definition locations in lib.d.ts, which can change between versions.

daviwil:
Fixing this now, will have an update pushed soon!

daviwil:
OK, I fixed the broken tests!  Turns out that there was another test that had an issue with TypeScript 2.6:

```
  6) TypeScriptService
       rootUri file:///foo/bar/
         textDocumentCompletion()
           produces completions for referenced symbols:
     Error: Debug Failure. False expression.
      at computePositionOfLineAndCharacter (node_modules\typescript\lib\typescript.js:5749:22)
```

This was caused by a [more strict assert](https://github.com/Microsoft/TypeScript/blob/v2.6.2/src/compiler/scanner.ts#L345) for text document positions added in the 2.6 series, it caught a bad character position in the test code.

By the way, any chance of a new version of the language server being published once this gets merged?  We'd love to include it in Atom's [ide-typescript](https://github.com/atom/ide-typescript) update later today!

package.json
```diff
@@ -51,7 +51,7 @@
     "rxjs": "^5.5.0",
     "semaphore-async-await": "^1.5.1",
     "string-similarity": "^1.1.0",
-    "typescript": "2.4.2",
+    "typescript": "~2.6.2",
```
felixfbecker: I don't know if we can guarantee that TS will still work on patch version upgrades. We also don't currently have a lockfile so this would be very risky.

daviwil: No problem!  I'll remove the tilde.

felixfbecker:
We release automatically with semantic-release, so yeah this would be released after merge

package.json
daviwil: Done!

src/typescript-service.ts
```diff
@@ -100,25 +100,25 @@ export interface Settings extends PluginSettings {
...
+    [`method`, CompletionItemKind.Method],
+    [`module`, CompletionItemKind.Module],
+    [`property`, CompletionItemKind.Property],
+    [`reference`, CompletionItemKind.Reference],
+    [`snippet`, CompletionItemKind.Snippet],
+    [`text`, CompletionItemKind.Text],
+    [`unit`, CompletionItemKind.Unit],
+    [`value`, CompletionItemKind.Value],
+    [`variable`, CompletionItemKind.Variable],
+])
```
felixfbecker: what was the reason to make this a Map, given it never changes?

daviwil: Using `constructor` as a object property name caused type system problems with `CompletionItemKind`, somehow it inferred that `Function` was a part of the union of numbers that `CompletionItemKind` is aliased to.  I couldn't find another way to escape the known keyword of `constructor` so using a map seemed to be the next best option.  Happy to change it back if you know of a better way!

felixfbecker: Ah, that makes total sense!

felixfbecker:
Thanks a lot!

daviwil:
🎉 No problem, and thank you too!

felixfbecker:
Btw, is there a reason why you didn't update to 2.7.2 (latest stable)?

daviwil:
Hmmm, perhaps I was under a mistaken assumption that even number patch releases were stable and odd numbers were unstable.  Happy to open another PR to get us up to 2.7.2 if you like!

daviwil:
Looks like a couple more tests are broken with the 2.7.2 update so it may be later today before I send the PR, but I'll get it done :)

daviwil:
Ahhh now I remember why we didn't use 2.7.2, it's because I started working on bumping the TS version last week before 2.7.1 was made stable, didn't think to check if there was a stable release this week :)  PR is imminent

PR #431, opened 2018-02-16
feat(typescript): upgrade to typescript 2.7.2

daviwil:
/cc @damieng

### DIFF
--- a/src/test/typescript-service-helpers.ts
+++ b/src/test/typescript-service-helpers.ts
@@ -382,7 +382,7 @@ export function describeTypeScriptService(
                             character: 8,
                         },
                     },
-                    contents: [{ language: 'typescript', value: 'import Foo' }, '**alias**'],
+                    contents: [{ language: 'typescript', value: 'class Foo\nimport Foo' }, '**alias**'],
                 })
             })
             specify('over keyword (non-null)', async function(this: TestContext & ITestCallbackContext): Promise<void> {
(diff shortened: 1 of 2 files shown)


src/test/typescript-service-helpers.ts
```diff
@@ -382,7 +382,7 @@ export function describeTypeScriptService(
                             character: 8,
                         },
                     },
-                    contents: [{ language: 'typescript', value: 'import Foo' }, '**alias**'],
+                    contents: [{ language: 'typescript', value: 'class Foo\nimport Foo' }, '**alias**'],
```
daviwil: Looks like hovers now give you more symbol info than just the import statement!

daviwil:
Anything I can/should do about the drop in code coverage?

felixfbecker:
Hmm, seems like `getNewLine()` is not called anymore by TS. Could you check if it is still part of the interface? If not, we can safely remove it

https://codecov.io/gh/sourcegraph/javascript-typescript-langserver/pull/431/changes#L95

daviwil:
Still part of the interface: https://github.com/Microsoft/TypeScript/blob/v2.7.2/src/services/types.ts#L163

Looks like there was a recent change to prioritize the use of `FormatCodeSettings` if it's provided before ever calling `getNewLine()`, so that could account for the lack of calls:

https://github.com/Microsoft/TypeScript/pull/21158/files#diff-183e2291cbb94d65ae0406354ce1ba7aR1262

I'll remove that function from the language service host and see what happens!

felixfbecker:
In that case just leave it in and not worry about the coverage

daviwil:
Ok, sounds good!

PR #450, opened 2018-04-23
fix: use safe-decode-uri-component

### LINKED ISSUE #422: Filenames with `%` breaks server
For a simplified situation, in a folder named sample-js, with two files,
- `test.js`
- `%test.js`

The server will quit with this error
```
/usr/local/lib/node_modules/javascript-typescript-langserver/lib/util.js:46
    const pathParts = parts.pathname.split('/').map(segment => encodeURIComponent(decodeURIComponent(segment)));
                                                                                  ^

URIError: URI malformed
    at decodeURIComponent (<anonymous>)
    at parts.pathname.split.map.segment (/usr/local/lib/node_modules/javascript-typescript-langserver/lib/util.js:46:83)
    at Array.map (<anonymous>)
    at Object.normalizeUri (/usr/local/lib/node_modules/javascript-typescript-langserver/lib/util.js:46:49)
    at Glob.globber.on (/usr/local/lib/node_modules/javascript-typescript-langserver/lib/fs.js:62:40)
    at Glob.emit (events.js:159:13)
    at Glob._emitMatch (/usr/local/lib/node_modules/javascript-typescript-langserver/node_modules/glob/glob.js:489:8)
    at Glob._processSimple2 (/usr/local/lib/node_modules/javascript-typescript-langserver/node_modules/glob/glob.js:705:8)
    at /usr/local/lib/node_modules/javascript-typescript-langserver/node_modules/glob/glob.js:676:10
    at Glob._stat2 (/usr/local/lib/node_modules/javascript-typescript-langserver/node_modules/glob/glob.js:789:10)
```

File names like `%` are used by vim as backup files and are not rare to encounter.

ramybenaroya:
This takes care of file names with `'%'` character. File names like
these can be automatically created by IDEs for backup.
Also, added a test for adding file with `'%'` character to `InMemoryFileSystem`.

Fixes #422

### DIFF
--- a/src/test/fs.test.ts
+++ b/src/test/fs.test.ts
@@ -2,7 +2,7 @@ import * as chai from 'chai'
 import chaiAsPromised = require('chai-as-promised')
 import * as fs from 'mz/fs'
 import * as path from 'path'
-import * as rimraf from 'rimraf'
+import rimraf from 'rimraf'
 import * as temp from 'temp'
 import { LocalFileSystem } from '../fs'
 import { path2uri } from '../util'
--- a/src/test/memfs.test.ts
+++ b/src/test/memfs.test.ts
@@ -18,6 +18,14 @@ describe('memfs.ts', () => {
                 sinon.assert.calledOnce(listener)
                 sinon.assert.calledWithExactly(listener, 'file:///foo/bar.txt', undefined)
             })
+            it('should add a URI with `%` character and emit an event', () => {
+                const listener = sinon.spy()
+                const fs = new InMemoryFileSystem('/')
+                fs.on('add', listener)
+                fs.add('file:///foo/%bar.txt')
+                sinon.assert.calledOnce(listener)
+                sinon.assert.calledWithExactly(listener, 'file:///foo/%bar.txt', undefined)
+            })
             it('should add content for a URI and emit an event', () => {
                 const listener = sinon.spy()
                 const fs = new InMemoryFileSystem('/')
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -9,6 +9,7 @@
     "sourceMap": true,
     "declaration": true,
     "strictFunctionTypes": false,
+    "esModuleInterop": true,
     "strictPropertyInitialization": false,
     "plugins": [
       {
--- a/package.json
+++ b/package.json
@@ -49,6 +49,7 @@
     "object-hash": "^1.1.8",
     "opentracing": "^0.14.0",
     "rxjs": "^5.5.0",
+    "safe-decode-uri-component": "^1.1.3",
     "semaphore-async-await": "^1.5.1",
     "string-similarity": "^1.1.0",
     "typescript": "2.7.2",
--- a/src/typings/safe-decode-uri-component.d.ts
+++ b/src/typings/safe-decode-uri-component.d.ts
@@ -1,0 +1,3 @@
+declare module 'safe-decode-uri-component' {
+    export default function safeDecodeURIComponent(uriComponent: string): string
+}
--- a/src/util.ts
+++ b/src/util.ts
@@ -1,5 +1,6 @@
 import { escapePathComponent } from 'fast-json-patch'
 import { Observable } from 'rxjs'
+import safeDecodeURIComponent from 'safe-decode-uri-component'
 import { compareTwoStrings } from 'string-similarity'
 import * as ts from 'typescript'
 import * as url from 'url'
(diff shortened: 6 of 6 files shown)


tsconfig.json
```diff
@@ -9,6 +9,7 @@
     "sourceMap": true,
     "declaration": true,
     "strictFunctionTypes": false,
+    "esModuleInterop": true,
```
ramybenaroya: Added this because `safe-decode-uri-component` is written in CommonJS. With this flag `safe-decode-uri-component` can be imported by:
```javascript
import safeDecodeURIComponent from 'safe-decode-uri-component'
```
rather than
```javascript
import * as safeDecodeURIComponent from 'safe-decode-uri-component'
```

src/test/memfs.test.ts
```diff
@@ -18,6 +18,14 @@ describe('memfs.ts', () => {
                 sinon.assert.calledOnce(listener)
                 sinon.assert.calledWithExactly(listener, 'file:///foo/bar.txt', undefined)
             })
+            it('should add a URI with `%` character and emit an event', () => {
```
ramybenaroya: This test would have been failed before this PR changes.
This is exactly the case #422 referres to.

src/test/fs.test.ts
```diff
@@ -2,7 +2,7 @@ import * as chai from 'chai'
 import chaiAsPromised = require('chai-as-promised')
 import * as fs from 'mz/fs'
 import * as path from 'path'
-import * as rimraf from 'rimraf'
+import rimraf from 'rimraf'
```
ramybenaroya: Fix build failure
https://travis-ci.org/sourcegraph/javascript-typescript-langserver/jobs/370047243#L1806

ramybenaroya:
Do you mean `'file:///foo/a%20bar.txt'`?
If so, yes it would pass.

felixfbecker:
I meant a file that is literally called `my%20file.txt`

ramybenaroya:
oh, I see what you mean.
You are right, my PR's purpose is only for fixing _invalid escaped_.
Generally, on runtime, at [this line](https://github.com/sourcegraph/javascript-typescript-langserver/pull/450/files#diff-8cd4968b81985f0efc2053eabc37db6dR47), `segment` argument will already be `my%2520file.txt` since [InMemoryFileSystem.add](https://github.com/sourcegraph/javascript-typescript-langserver/blob/master/src/memfs.ts#L77)'s input is a URI, meaning a valid escaped URL.

I made this PR to fix #422 (where for some reason paths are not escaped properly) in order to prevent a fatal exception.

felixfbecker:
Okay, in that case I believe the issue should be fixed properly.

ramybenaroya:
Do you want to close the PR?
I am ready to help with a proper fix if you can guide me a bit since this is my first contribution in this project.
What do you have in mind for a proper fix?

autozimu:
Sorry for bluntly chime in.

#422 is not caused by invalid encoding.

The file name is literally `%test.js`, which is used by vim as backup file by default (vaguely remember emacs does the same thing).

As I also commented under #422, language client didn't send this path. It is picked up by language server when trying to build index.

With that said, this should resolve #422, though it only fixes the consequeue but not the cause.

I'd love to see #422 got resolved, but a concern from this project's perspective is that how much compatibility this new `safe-decode-uri-component` can provide as a replacement of `decodeURIComponent `, when it comes to decoding valid uris.

ramybenaroya:
@autozimu Thanks for the explanation.
I am closing this for the following reasons:
1.`safe-decode-uri-component`is probably not a good candidate for replacing native `decodeURIComponent` since it is not widely used and doesn't meet this project's standards.
2. A better solution will be to fix the index building process in order that `%test.js` will be encoded to `%25test.js`.
3. Fixing the consequence is never good enough.

@felixfbecker @autozimu Thanks for your input and attention.

felixfbecker:
For context, the reason why the code calls `encodeURIComponent(decodeURIComponent(segment))` is so that files names have a canonical name and can be used as map keys. Otherwise you can end up with the client sending URIs for certain files with a different URI than what the language server came up with and you get duplicate entries in Maps. For example, a client might encode a space as `+`, but the language server used `%20`. Or the language server chose to escape some special character that the client didn't.

Now the bug here seems to be indeed that the URI for `%test.js` at that point in the code is not properly encoded to `%25test.js`. I would set a conditional breakpoint inside the function (or an `if` with `console.trace()`) and see where the value comes from.

autozimu:
The call comes from https://github.com/sourcegraph/javascript-typescript-langserver/blob/312e432215311ae4604168d6fc6d0c7cfb2301bb/src/fs.ts#L81

The result come from glob should already be valid paths.

For the concern that the path might have other aliases other than its canonical form. I believe it can be done by querying from filesystem, https://github.com/autozimu/LanguageClient-neovim/blob/84445be4a3bf7f67547abad205c6c46eb289844d/src/languageclient.rs#L1619 It won't cover all cases, but should be enough for general cases.

felixfbecker:
Oh, seems like all it needs it to encode `file` in that line! When constructing the URI from file paths, it needs to be encoded.

> For the concern that the path might have other aliases other than its canonical form. I believe it can be done by querying from filesystem

That won't work because not all files the language server works with are available on the file system.

PR #462, opened 2018-05-18
Add missing project method for plugins

soywod:
I'm using [tslint-language-service](https://github.com/angelozerr/tslint-language-service/issues), and this plugin doesn't work with javascript-typescript-langserver because there is some missing properties/methods in your PluginCreateInfo. So I just added the `getCurrentDirectory`.

### DIFF
--- a/src/plugins.ts
+++ b/src/plugins.ts
@@ -40,7 +40,7 @@ export interface PluginCreateInfo {
  * The portion of tsserver's Project API exposed to plugins
  */
 export interface Project {
-    getCurrentDirectory: () => string,
+    getCurrentDirectory: () => string
     projectService: {
         logger: Logger
     }
--- a/src/plugins.ts
+++ b/src/plugins.ts
@@ -40,6 +40,7 @@ export interface PluginCreateInfo {
  * The portion of tsserver's Project API exposed to plugins
  */
 export interface Project {
+    getCurrentDirectory: () => string,
     projectService: {
         logger: Logger
     }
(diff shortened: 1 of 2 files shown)


src/plugins.ts
```diff
@@ -40,6 +40,7 @@ export interface PluginCreateInfo {
  * The portion of tsserver's Project API exposed to plugins
  */
 export interface Project {
+    getCurrentDirectory: () => string
```
felixfbecker: Style nit, could this just be a method rather than an arrow function?

getCurrentDirectory(): string

src/project-manager.ts
```diff
@@ -385,7 +385,11 @@ export class ProjectConfiguration {
 
             const info: PluginCreateInfo = {
                 config: configEntry,
-                project: { projectService: { logger: this.logger } }, // TODO: may need more support
+                project: {
+                    // TODO: may need more support
+                    getCurrentDirectory: this.getHost().getCurrentDirectory.bind(this),
```
felixfbecker: bind() is not type safe, prefer wrapping in a function manually

src/plugins.ts
soywod: Yes you right, I update this

src/project-manager.ts
soywod: :heavy_check_mark:

src/plugins.ts
soywod: :heavy_check_mark:

PR #472, opened 2018-06-13
Do not aggressively fetch all workspace files when using the local filesystem

keyboardDrummer:


Kronuz:
This definitely fixes the warmup delay in #477

felixfbecker:
> The reason is that this LSP server allows for multiple projects to exist inside of the workspace, so it must scan to workspace to find these projects. However, the LSP protocol does not require this. You may assume the workspace root is the root of a single project. For handling multiple projects with one LSP server, a client can use the recently introduced workspace folders feature. We should change this LSP server so it uses that instead of scanning the root directory.

I don't think we can do that - a user doesn't necessarily have each project in its own workspace folder, but may add a monorepo as a single workspace folder. For Sourcegraph, we don't know what "defines" a project in a repo, so every connection passes the repo root as the root URI. Only the langserver has the knowledge to know where to draw language-specific project boundaries - but we have to work with the limitations of `LanguageServiceHost` being sync. If a client wants to reduce the number of files scanned, it is free to spawn one language server instance for a subfolder that only contains one project (or use `workspaceFolders` if we add support for it).

keyboardDrummer:
> I have trouble understanding the code architecture after this change. Before I believe the code followed SRP very well: InMemoryFileSystem was the synchronous store that TS would talk to. It was populated by FileSystemUpdater, which synced files to it from any implementation of the async FileSystem interface - either RemoteFileSystem or LocalFileSystem.

The core of the change is that the `files` property from `InMemoryFileSystem` has been moved to `RemoteFileSystem`. Previously `InMemoryFileSystem` did these two things:
1. ensure that files in the filesystem can be served synchronously
1. overlay the file system with files that aren't saved yet.

Responsibility (1) has been moved to the `FileSystem` interface, which is why `InMemoryFileSystem` has been renamed to `OverlayFileSystem`. The move is necessary because it allows the `LocalFileSystem` to decide to serve files synchronously straight from disk instead of storing them. Microsoft's `LanguageService` back-end already caches all file contents it requests, so performance wise there is no need for the `LanguageServiceHost` to do the same.

`MapFileSystem` (which is used for testing) and `RemoteFileSystem` both require storing files in memory to implement responsibility (1), so that feature has been moved to a common base class called `InMemoryFileSystem`. The class `InMemoryFileSystem` should be located in `fs.ts`, but since I wanted to make the diff shorter and the `files` property was previously in `memfs.ts`, I put `InMemoryFileSystem` there.

---

An interesting question is what is faster:
1. (this PR): intertwine the synchronous reading of relevant files with running the Microsoft back-end
1. (master): first read all files in the workspace folder that may be relevant asynchronously and then run the Microsoft back-end

For packages with a directory that contains many files that don't relate to the analysis, which is the case for our package, option (1) works while (2) never completes initializing. However, maybe for packages that only contain relevant files (2) could be faster.

--- 
> For Sourcegraph, we don't know what "defines" a project in a repo, so every connection passes the repo root as the root URI.

I see, so you definitely need the LSP server to support finding all projects inside the root folder. There could still be a flag though for indicating that the root folder (or workspaceFolders if that feature is used) is/are the roots of the projects.

src/fs.ts
```diff
@@ -225,3 +417,4 @@ export class FileSystemUpdater {
         this.structureFetch = undefined
     }
 }
+>>>>>>> Stashed changes
```
ccope: merge conflict marker

src/fs.ts
```diff
@@ -51,8 +222,11 @@ export class RemoteFileSystem implements FileSystem {
             .map(textDocument => textDocument.text)
     }
 }
+<<<<<<< Updated upstream
```
ccope: Merge conflict marker

ccope:
Actually I just had to revert the last commit in the branch to get a functional build.

PR #475, opened 2018-06-16
feat: Add goto type definition.

teasp00n:
Adds support to go directly to the type definition for the symbol under the cursor.

### DIFF
--- a/src/typescript-service.ts
+++ b/src/typescript-service.ts
@@ -297,6 +297,7 @@ export class TypeScriptService {
                     triggerCharacters: ['(', ','],
                 },
                 definitionProvider: true,
+                typeDefinitionProvider: true,
                 referencesProvider: true,
                 documentSymbolProvider: true,
                 workspaceSymbolProvider: true,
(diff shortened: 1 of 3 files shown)


src/typescript-service.ts
```diff
@@ -417,6 +417,68 @@ export class TypeScriptService {
...
+                    return {
+                        uri: locationUri(definition.fileName),
+                        range: {
+                            start,
+                            end,
+                        },
+                    }
+                })
+            })
+    }
```
felixfbecker: It seems like a lot of this implementation is copy-pasted except one line, would it maybe be possible to share to most the code?

teasp00n:
> You also need to announce the capability in the InitializeResult.

Not sure what I need to do for this. `ServerCapabilities` appears to just extend `vscode.ServerCapabilities` with functions specific to this language server. The go to type is part of the spec and should be advertised already as part of `vscode.ServerCapabilities` or have I misunderstood what you mean?

felixfbecker:
I added the capability

PR #478, opened 2018-06-19
feat: add textDocument/implementation support

felixfbecker:


src/typescript-service.ts
```diff
@@ -453,6 +454,70 @@ export class TypeScriptService {
             .startWith({ op: 'add', path: '', value: [] })
     }
 
+    /**
+     * The goto type definition request is sent from the client to the server to resolve the type
```
teasp00n: Looks like the original function doc is still here, probably want to update that :)

teasp00n:
@felixfbecker Is this blocked on something?

felixfbecker:
The only thing it's blocked on is that for some reason TypeScript does not return any results in the tests even though I would expect it to. Can't really merge if it always returns no result...

PR #480, opened 2018-06-20
fix(imports): stop using import assignment (for ECMAScript modules)

Kronuz:
When trying to compile as ECMAScript modules, errors about import assignment don't allow compilation. This pull request fixes that by changing such to use `import * as XXX from YYY` instead.

### DIFF
--- a/src/typescript-service.ts
+++ b/src/typescript-service.ts
@@ -2,7 +2,7 @@ import { Operation } from 'fast-json-patch'
 import iterate from 'iterare'
 import { castArray, merge, omit } from 'lodash'
 import { toPairs } from 'lodash'
-import hashObject = require('object-hash')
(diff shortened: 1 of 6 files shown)


src/typescript-service.ts
```diff
@@ -2,7 +2,7 @@ import { Operation } from 'fast-json-patch'
 import iterate from 'iterare'
 import { castArray, merge, omit } from 'lodash'
 import { toPairs } from 'lodash'
-import hashObject = require('object-hash')
+import * as hashObject from 'object-hash'
```
felixfbecker: Using `* as` for functions that are exported with `module.exports` is not semantically correct - according to the ES6 spec `hashObject` would be a frozen module record object, not a function.

The right solution for this would be to use the new `esModuleInterop` flag and import the default import.

PR #492, opened 2018-07-24
symbols: do not limit symbols found from config

### LINKED ISSUE #491: Symbols limiting for workspace/symbol works incorrectly
### Overview

Currently when reading workspace symbols in config (see `_getSymbolsInConfig` method), amount of symbols limited to 100.
https://github.com/sourcegraph/javascript-typescript-langserver/blob/d107424f1f964c4cb3b6061570890639fd69b9e6/src/typescript-service.ts#L1681
Also the results is filtered latter using check `item.fileName.includes('/node_modules/')`
https://github.com/sourcegraph/javascript-typescript-langserver/blob/d107424f1f964c4cb3b6061570890639fd69b9e6/src/typescript-service.ts#L1684
### Problem

Assume you type character `m` to get symbols for it.
It will return 100 symbols but 99 of them could be from any of `node_modules` packages.
Filtering out leaves us with 1 symbol returned.

After that you're narrowing down your search query to `my`.
It will return again 100 symbols but in this case as you're looking for symbols in your project and already more specific in your query it could return for example only 60 results from `node_modules` and the rest 40 from your package.
Filtering out leaves us with 40 symbol returned.

So as a result it causes the confusion as when I have broad search term it leaves me with lower amount of results comparing with more specific search term.

### Proposed Solution

Here is the PR for the proposed solution: https://github.com/sourcegraph/javascript-typescript-langserver/issues/491

Proposed solution is not to include `maxResultCount` into to `getNavigateToItems` to TS server.
Results are filtered later in `workspaceSymbol`.

#### Performance
Current implementation of `getNavigateToItems` in TS server (https://github.com/Microsoft/TypeScript/blob/64555aa6a94879f8fee7c0718e688a09cc8d4073/src/services/navigateTo.ts#L11) anyway goes from all symbols in source files and slicing array symbols later.

The only changes is that `createNavigateToItem` function would be on larger amount of items.
https://github.com/Microsoft/TypeScript/blob/64555aa6a94879f8fee7c0718e688a09cc8d4073/src/services/navigateTo.ts#L30

andrewnester:
limiting symbols found from config could lead to confusion when broad search term returns less results that more specific one.

Fixes #491

### DIFF
--- a/src/typescript-service.ts
+++ b/src/typescript-service.ts
@@ -1676,9 +1676,8 @@ export class TypeScriptService {
 
             if (typeof query === 'string') {
                 // Query by text query
-                // Limit the amount of symbols searched for text queries
                 return (
-                    Observable.from(config.getService().getNavigateToItems(query, 100, undefined, false))
+                    Observable.from(config.getService().getNavigateToItems(query))
                         // Exclude dependencies and standard library
                         .filter(
                             item => !isTypeScriptLibrary(item.fileName) && !item.fileName.includes('/node_modules/')


andrewnester:
@felixbecker yeah, I could agree with this point, maybe it's better to have higher limit then no limit at all.

As to use case I described it in this issue https://github.com/sourcegraph/javascript-typescript-langserver/issues/491

felixfbecker:
Cool, if it still has a limit applied at the end and the TS limit implementation does nothing but slice the final array I am in favor of this change.

andrewnester:
wow, that was quick :) thanks @felixfbecker !

Issue #520, opened 2018-09-13
operand checked twice in or statement.


francisschmaltz:
On line 1011 of `src/typescript-service.ts`
```ts
if (!params.query.package || !params.query.package) {
     return [symbol]
}
```
should be 
```ts
if (!params.query.package) {
     return [symbol]
}
```

felixfbecker:
Wanna do a PR?

nicksnyder:
Yeah, just edit the code @francisschmaltz  :)

PR #521, opened 2018-09-13
Correct operand checked twice in or statement. #520

francisschmaltz:
Correct operand checked twice in or statement. #520


felixfbecker:
@francisschmaltz for your next PR, please make sure to follow the commit message guidelines at https://conventionalcommits.org/
Not critical for this one because it will just not release this commit and the commit is not important

PR #523, opened 2018-09-24
Set correct range field for errors that have a length of zero

jzashch:
The server is sending invalid diagnostics line if error length is zero.

### DIFF
--- a/src/diagnostics.ts
+++ b/src/diagnostics.ts
@@ -4,14 +4,11 @@ import { Diagnostic, DiagnosticSeverity, Range } from 'vscode-languageserver'
 /**
  * Converts a TypeScript Diagnostic to an LSP Diagnostic
  */
-export function convertTsDiagnostic(diagnostic: ts.Diagnostic): Diagnostic {
+export function convertTsDiagnostic(diagnostic: ts.DiagnosticWithLocation): Diagnostic {
     const text = ts.flattenDiagnosticMessageText(diagnostic.messageText, '\n')
-    let range: Range = { start: { character: 0, line: 0 }, end: { character: 0, line: 0 } }
-    if (diagnostic.file && diagnostic.start && diagnostic.length) {
-        range = {
-            start: diagnostic.file.getLineAndCharacterOfPosition(diagnostic.start),
-            end: diagnostic.file.getLineAndCharacterOfPosition(diagnostic.start + diagnostic.length),
-        }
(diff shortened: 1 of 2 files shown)


felixfbecker:
Could you explain how the previous range was invalid? I don't see in the spec where it says that empty ranges are invalid

src/diagnostics.ts
```diff
@@ -4,14 +4,11 @@ import { Diagnostic, DiagnosticSeverity, Range } from 'vscode-languageserver'
 /**
  * Converts a TypeScript Diagnostic to an LSP Diagnostic
  */
-export function convertTsDiagnostic(diagnostic: ts.Diagnostic): Diagnostic {
+export function convertTsDiagnostic(diagnostic: ts.DiagnosticWithLocation): Diagnostic {
     const text = ts.flattenDiagnosticMessageText(diagnostic.messageText, '\n')
-    let range: Range = { start: { character: 0, line: 0 }, end: { character: 0, line: 0 } }
-    if (diagnostic.file && diagnostic.start && diagnostic.length) {
```
jzashch: Here you first set the initial lines for start and end to zero. 
Then you check if file and start and **length** are not _null/undefined/**0**_. 
In case the file has a syntax error, length sent from TS server is zero. So the value for range doesn't get recalculated. 
Something like
```
class Test {

}

const test = new Test();

test.
```
The last line should give you the syntax error line **Identifier expected**, but the range will be invalid - it'll start at line 0 instead of line 7.

src/typescript-service.ts
```diff
@@ -1645,6 +1645,10 @@ export class TypeScriptService {
                     .filter((e): e is ts.DiagnosticWithLocation => !!e.file)
             )
         const diagnostics = iterate(tsDiagnostics)
+            // TS can report diagnostics without a file and range in some cases
+            // These cannot be represented as LSP Diagnostics since the range and URI is required
+            // https://github.com/Microsoft/TypeScript/issues/15666
+            .filter(diagnostic => !!diagnostic.file && diagnostic.start !== undefined && diagnostic.length !== undefined)
```
felixfbecker: So doesn't this filter out diagnostics without a range, whereas before they were reported as diagnostics with a range of 0:0-0:0?

jzashch: I see your point, so I moved the check. We still need to separate length == 0 from length == undefined in it.

PR #553, opened 2019-01-07
fix(deps): move commitlint packages to devDependencies

ybiquitous:
`commitlint` ~~and `chai`~~ packages seem to be used only on development.

- `commitlint`: used on pre commit
- ~~`chai`: used on test~~

This PR removes these extra dependencies from runtime dependencies.

### DIFF
--- a/package.json
+++ b/package.json
@@ -41,8 +41,6 @@
     ]
   },
   "dependencies": {
-    "@commitlint/cli": "^7.1.2",
-    "@commitlint/config-conventional": "^7.1.2",
     "chai": "^4.0.1",
     "chai-as-promised": "^7.0.0",
     "chalk": "^2.2.0",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -78,6 +76,8 @@
     "@types/rimraf": "2.0.2",
     "@types/sinon": "4.3.1",
     "@types/temp": "0.8.32",
+    "chai": "^4.0.1",
+    "chai-as-promised": "^7.0.0",
     "commitizen": "^2.9.6",
```
ybiquitous: [note] It looks `commitizen` should be removed... (_unrelated to this PR_)

ybiquitous:
Thanks for your review! I rewrote the commit, the PR title & message.

ybiquitous:
I have a question. Is there no need to include also `mocha` & `sinon` into `dependencies`?

felixfbecker:
I guess yes

ybiquitous:
Thank you!

PR #571, opened 2019-03-27
fix(deps): move chai and chai-as-promised to dev-dependencies

rhysd:
While checking `npm list -g`, I noticed assertion libraries for unit tests were installed on my local. They should be in `dev-dependencies` so I moved them.


felixfbecker:
The test suite is published to npm and can be used by extensions of the language server, so we need to have these as true dependencies.

